PR #28, opened 2026-03-17
fix(node): remove references to unexported findRosterCards and loadPreset

parametrization:
## Summary
- Removed `findRosterCards` and `loadPreset` imports from test file (no longer exported after PR #22)
- Removed 3 `findRosterCards` tests that referenced the unexported function
- Replaced 2 `loadPreset` tests with `listPresets()`-based equivalents
- All 69 tests pass

## Test plan
- [x] `npx vitest run` passes with 69/69 tests green
- [x] No references to `findRosterCards` or `loadPreset` remain in test file

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/node/tests/cli.test.ts
+++ b/node/tests/cli.test.ts
@@ -28,8 +28,6 @@ import {
   extractField,
   replaceField,
   safeName,
-  findRosterCards,
-  loadPreset,
   listPresets,
   bootstrap,
   addMember,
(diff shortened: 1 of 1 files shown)


parametrization (commented):
Review: I've taken a different approach on this in PR #30 — instead of removing these tests, I exported `loadPreset` and added `findRosterCards` back to `bootstrap.ts`. Both functions are useful for the test suite and potentially for consumers of the library.

That said, if this needs to merge to unblock main right now, the approach is fine — the tests can be re-added when the exports land from PR #30. Either way works.

One note: the `loadPreset` test replacement (`listPresets().find(...)`) doesn't test the error path for unknown presets, which the original `loadPreset("nonexistent")` test covered. Consider adding:
```ts
it("should not find nonexistent preset", () => {
  const presets = listPresets();
  expect(presets.find(p => p.name === "nonexistent")).toBeUndefined();
});
```

PR #112, opened 2026-07-05
fix(trust-signals): phase-aware branch.integration template (#100)

parametrization:
## Summary
Makes the `branch.integration` template **phase-aware** — the real treatment behind issue #100 (the Manager's config-value edit was only a stopgap).

`trust_signals._integration_base` resolved the merged-PR base via `template.format(wave=wave)`, so a `{phase}` token raised `KeyError` and fell back to the raw literal template — silently scoping the retro to zero PRs for phase-namespaced projects (`deployments/phase{phase}/wave-{wave}`).

## Changes
- **`framework/assets/lib/trust_signals.py`**
  - `_integration_base(cfg, wave, phase=None)` now substitutes **both** `{phase}` and `{wave}` through a partial format map (`_PartialTokens`) that leaves unknown/unresolved tokens literal instead of raising — mirrors the shell skills' `sed` behaviour.
  - New `_phase_for_wave(wave, status_path)` resolves phase from `wave_<id>_phase` in the state file (fail-open → None).
  - `merged_prs` wires the resolved phase into the base derivation.
- **retro / wave-retro skill guidance** — updated the `branch.integration` substitution note to include `{phase}` (canonical `framework/assets/skills/` **and** live `.claude/skills/` copies kept in sync — dual-deploy).
- **`framework/tests/test_trust_signals.py`** (new, 13 tests) — both-token substitution, safe unresolved-token fallback, generic wave-only default, literal named branch, and state-driven phase resolution end-to-end.

## Consumer audit (phase-token handling verified)
- `trust_signals._integration_base` — **fixed here**.
- `wave-start`, `wave-audit` skills — already phase-aware (`sed s/{phase}/…/; s/{wave}/…/`), default `deployments/phase{phase}/wave-{wave}`.
- `block_squash_wave_merge.py` hook — already maps `{phase}`/`{wave}` → `\d+`.
- Schema (`framework.config.schema.json`) already documents both tokens.
- Generic embedded defaults (`_framework_config.py`, schema, `bootstrap.py`) intentionally left as `deployments/wave-{wave}` — the phase-less default is correct for fresh non-phased projects; the substitution logic handles `{phase}` whenever a project opts into it.

## Verification
- `pytest framework/tests/` → **270 passed** (13 new).

## Risks / decisions
- Unresolved `{phase}` (no state / generic project) degrades to a literal `{phase}` in the branch name rather than crashing — a visible, debuggable fail vs. the previous silent zero-PR result.
- Did not touch the archived intake candidate copy (`intake/.../candidates/lib/trust_signals.py`) — it is a frozen snapshot, not a live runtime path.

Closes #100

### DIFF
--- a/.claude/skills/retro/SKILL.md
+++ b/.claude/skills/retro/SKILL.md
@@ -33,8 +33,8 @@ LIB="$REPO_ROOT/.claude/lib"
 
 Read `current_wave` (call it `{W}`) from the state output, or ask the user. The wave label
 is `labels.wave` with `{id}` → `{W}`. The PR base branch is `branch.integration` (with
-`{wave}` → `{W}`) when the wave's merge model is `wave-branch`, otherwise the default
-branch.
+`{phase}` → the wave's phase and `{wave}` → `{W}`) when the wave's merge model is
+`wave-branch`, otherwise the default branch.
 
 ### 1. Collect progress data
 
--- a/.claude/skills/wave-retro/SKILL.md
+++ b/.claude/skills/wave-retro/SKILL.md
@@ -49,7 +49,7 @@ The PR base depends on the wave's merge model:
 
 ```bash
 MM="$(python3 "$LIB/lifecycle.py" merge-model get "{W}" 2>/dev/null || echo direct-to-main)"
-# wave-branch     → BASE = branch.integration template with {wave} → {W}
+# wave-branch     → BASE = branch.integration template ({phase} → wave's phase, {wave} → {W})
 # direct-to-main  → BASE = $DEFAULT_BRANCH, additionally filtered by the wave label
 gh pr list --state merged --base "{BASE}" --json number,title,author,body,mergedAt,reviews
 ```
--- a/framework/assets/lib/trust_signals.py
+++ b/framework/assets/lib/trust_signals.py
@@ -335,21 +335,63 @@ def _resolve_repos(cfg) -> list[str]:
     return [nwo] if nwo else []
 
 
-def _integration_base(cfg, wave: str) -> str:
+class _PartialTokens(dict):
+    """Format map that leaves unknown ``{token}`` placeholders literal.
+
+    Mirrors the shell skills' ``sed "s/{phase}/…/; s/{wave}/…/"`` behaviour:
+    known tokens are substituted, anything else is passed through verbatim
+    instead of raising ``KeyError``. Keeps template rendering fail-safe.
+    """
+
+    def __missing__(self, key: str) -> str:  # noqa: D105
+        return "{" + key + "}"
+
+
+def _render_branch_template(template: str, **tokens) -> str:
+    """Substitute *tokens* into *template*, leaving unknown placeholders literal."""
+    try:
+        return str(template).format_map(_PartialTokens(**tokens))
+    except (KeyError, IndexError, ValueError):
+        return str(template)
+
+
+def _phase_for_wave(wave: str, status_path: Path | None) -> str | int | None:
+    """The phase a wave belongs to (``wave_<id>_phase`` in state), or None.
+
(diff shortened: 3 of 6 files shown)


parametrization:
Requestor: Paloma Gupta
Requestee: Ibrahim El-Amin
RequestOrReplied: Request

**Charter review of PR #112 (issue #100, phase-aware `branch.integration`).** I did not author this PR. Verdict: **CHANGES-REQUESTED** — one blocking coordination item, which I co-own (I authored the colliding #113).

The core fix is correct and well-tested. `_integration_base` now substitutes `{phase}` + `{wave}` through `_PartialTokens.format_map`, leaving unresolved tokens literal instead of raising `KeyError`, and `_phase_for_wave` resolves phase from state. I verified the load-bearing premise: `wave_<id>_phase` is really stamped by the wave allocator (`lifecycle.py:359` and `:427`), so this reads live state and genuinely closes the silent zero-PR bug — not dead code. Dual-deploy checks out: there is no live `.claude/lib/trust_signals.py` to sync, both `retro`/`wave-retro` SKILL copies are updated in lockstep, and the frozen `intake/.../candidates/` snapshot is correctly left alone. Scope/altitude is right — the generic `deployments/wave-{wave}` defaults in `_framework_config.py`/schema/`bootstrap.py` are correctly kept phase-less. The consumer audit in the PR body is accurate.

Must-fix:
1. **Add/add conflict with #113 on `framework/tests/test_trust_signals.py`.** Both PRs create that exact path fresh (the base branch `deployments/phase4/wave-1` has no such file — only the unrelated `intake/.../candidates/lib/tests/` one), with fully disjoint content. Whichever merges second hits a guaranteed add/add conflict and cannot merge clean. Good news: the `trust_signals.py` **source** hunks do NOT overlap — #113 touches the module docstring, the `_FIELD_RE`/`Verdict`/`parse_verdicts` region (~L108-194) and the `_bucket` loop (~L514); #112 touches `_integration_base`/`_phase_for_wave` (~L252) and the `merged_prs` wiring (~L391), and adds only new symbols (`_PartialTokens`, `_render_branch_template`, `_phase_for_wave`) with no name clash against #113's (`_MUST_FIX_LABEL_RE`, `_has_must_fix_items`, the re-signatured `_is_changes_requested`). So a git 3-way on the `.py` merges cleanly in either order — **only the test file needs reconciliation.** Resolution (I will own the #113 side): land **#113 first** (larger 25-test corpus + the header superset — it puts both the `assets/hooks` and `assets/lib` dirs on `sys.path`, a superset of #112's lib-only header), then rebase #112 and **append its 13 template tests into the now-existing `test_trust_signals.py`**, keeping #113's import header. Acceptable lower-effort fallback if you'd rather not combine: split the filenames (`test_trust_signals_templates.py` here, `test_trust_signals_verdicts.py` for #113) — order-independent, zero merge work. Either way, do not let the second merge auto-resolve the add/add.

Tech-debt:
1. **`_phase_for_wave` duplicates `lifecycle.phase_of` (`lifecycle.py:234`).** Both key on `wave_<id>_phase`; `lifecycle.phase_of` gates on `isinstance(direct, int)` while yours returns any non-None value. This is consistent with the existing `_kickoff_ts` idiom (self-contained fail-open state read, no `lifecycle` import), so it is not blocking — but a shared resolver would prevent the two from drifting (e.g. if phase ever becomes a string, they diverge). File as a follow-up if you agree.
2. **No test exercises `merged_prs` passing the resolved phase through** — the end-to-end test composes `_phase_for_wave` + `_integration_base` by hand rather than asserting the `merged_prs` wiring line. Thin, but it is the actual integration point. Optional.
3. Minor: malformed-JSON state is caught by the `except (OSError, json.JSONDecodeError, ValueError)` but only the missing-file path is unit-tested; a one-line bad-JSON case would pin the fail-open contract.

PR #115, opened 2026-07-05
feat(hooks): validate_review_comment_format — enforce charter verdict-comment grammar (#111)

parametrization:
## Summary
- Adds `validate_review_comment_format` — a PreToolUse gate on `gh pr comment` / `gh issue comment` that blocks a comment whose body **attempts** the charter verdict-comment header but is malformed (missing/empty field, or an unknown `RequestOrReplied` token). Free-form comments with no header pass untouched.
- Makes the charter's review-comment shape reliable at write time so `trust_signals.py` stops silently reading review signals as **zero** — the failure #98 fixes on the parsing side. **This PR defines the vocabulary #98 parses.**
- Narrow pull from the deferred review-gate tranche (NOT the full N-reviewer gate).

## Canonical verdict-comment grammar (the vocab #98 must align to)
Documented in `.claude/team/charter/issues.md` § Verdict-Comment Grammar (and the canonical `framework/assets/` copy):

| Element | Label | Allowed values | Meaning |
|---|---|---|---|
| Header | `Requestor:` | name (bare or `**bold**`) | comment author / reviewer |
| Header | `Requestee:` | name, or `N/A` | addressee / PR author |
| Header | `RequestOrReplied:` | **`Request`** \| **`Replied`** | posting a verdict / replying |
| Body | `Must-fix:` | enumerated items, or `None` | enumerated ⇒ changes requested (blocks merge); `None` ⇒ clean |
| Body | `Tech-debt:` | items, or `None` | non-blocking |

Severity lives in the **body** (`Must-fix:`), never in the verdict token (no `Approved` / `ChangesRequested`). Both bare and bold header forms accepted.

**Coupling with #98:** the hook's `_STRICT_FIELD_RE` is byte-identical to `trust_signals._FIELD_RE`, and `test_field_regexes_match_trust_signals` pins that equality so the gate and the scorer cannot drift. Exported constants `HEADER_FIELDS` / `VERDICT_STATES` / `SEVERITY_MARKERS` are the machine vocab for #98 to bind to.

## Changes
- `framework/assets/hooks/validate_review_comment_format.py` — the hook (pure grammar, fail-open, stdlib + shared `_shell_parse`/`_framework_log`).
- Wired into `hooks.pre_bash` in all synced copies: schema, `_framework_config._DEFAULTS`, `bootstrap._schema_defaults()`, example config, and live dogfood config.
[...]

PR #121, opened 2026-07-05
fix(trust_signals): render {wave} as phase-local ordinal, not global seq (#117)

parametrization:
## Issue
Closes #117.

`trust_signals._integration_base` filled the `{wave}` token with the **global wave sequence** id, but a phase-namespaced integration branch is named by the wave's **phase-local ordinal**. Scoring Phase 4 Wave 1 (global wave 2 / phase 4 / ordinal 1) rendered `deployments/phase4/wave-2`, so `gh pr list --base` matched **0 PRs** and the retro scored every engineer empty. This is the sibling of #100 one layer down: #100 made `{phase}` substitutable; `{wave}` was still fed the wrong number.

## Fix
- Add `_wave_ordinal_for_wave(wave, status_path)` mirroring `_phase_for_wave`, reading `wave_<id>_phase_ordinal` from state (stamped by `lifecycle.py` alongside `wave_<id>_phase`).
- `_integration_base` now resolves `{wave}` to the phase-local ordinal when supplied, **falling back to the global wave id** for generic (non-phased) projects that stamp no ordinal.
- `merged_prs` passes the ordinal alongside the phase.

## Verification
- `ruff check framework/` clean; `pytest framework/tests/test_trust_signals.py` → 45 passed.
- New tests: `_wave_ordinal_for_wave` state reader (+ missing-key / no-path / unreadable), phase-ordinal substitution in `_integration_base`, generic fallback, and the end-to-end global-2 → `deployments/phase4/wave-1` wiring.
- **Live proof (no override):** `python3 framework/assets/lib/trust_signals.py extract 2 --status .claude/state.json` from a `deployments/phase4/wave-2` checkout now resolves `deployments/phase4/wave-1` and finds the 4 Wave 1 PRs (#112–#115). Before the fix it found 0. Wave 1 self-scores.

## Dual-deploy (#116)
Canonical `framework/assets/lib/trust_signals.py` edited. There is no live `.claude/lib/trust_signals.py` today (skills fall back to the asset copy), so **no reinstall is needed now**; once #116's reinstall mechanism lands, this change must be reinstalled onto the repo.

## Sequencing vs #119
Independent PR; the diffs are disjoint — this PR touches `_wave_ordinal_for_wave` / `_integration_base` / `merged_prs` (+ tests in the integration-branch section); #119 touches `extract_signals` + new roster helpers (+ tests in a separate section). Verified conflict-free by a local test-merge of both branches: both files auto-merge and the combined `test_trust_signals` suite is 52 green. Both branch off `deployments/phase4/wave-2` and merge in either order.

### DIFF
--- a/framework/assets/lib/trust_signals.py
+++ b/framework/assets/lib/trust_signals.py
@@ -373,7 +373,34 @@ def _phase_for_wave(wave: str, status_path: Path | None) -> str | int | None:
     return val if val is not None else None
 
 
-def _integration_base(cfg, wave: str, phase: str | int | None = None) -> str:
+def _wave_ordinal_for_wave(wave: str, status_path: Path | None) -> str | int | None:
+    """The phase-local ordinal of a wave (``wave_<id>_phase_ordinal``), or None.
+
+    PROJECT-COUPLING: keyed on a state-file entry named
+    ``wave_<id>_phase_ordinal`` stamped by the wave allocator alongside
+    ``wave_<id>_phase`` (see ``lifecycle.py``). A phase-namespaced project numbers
+    its integration branches by the position of the wave *within its phase*
+    (``deployments/phase4/wave-1`` is the first wave of phase 4), not by the
+    monotonic global wave id, so the ``{wave}`` token must resolve to this
+    ordinal. Absent in a generic (non-phased) project — or when the state file is
+    missing/unreadable — → None, and the caller falls back to the global wave id.
+    """
+    if status_path is None:
+        return None
+    try:
+        data = json.loads(Path(status_path).read_text())
+    except (OSError, json.JSONDecodeError, ValueError):
+        return None
+    val = data.get(f"wave_{wave}_phase_ordinal")
+    return val if val is not None else None
+
+
+def _integration_base(
+    cfg,
+    wave: str,
+    phase: str | int | None = None,
+    wave_ordinal: str | int | None = None,
+) -> str:
     """The merged-PR base branch for *wave*, from the integration-branch template.
 
     PROJECT-COUPLING: the original tool scoped the base to a phase+wave path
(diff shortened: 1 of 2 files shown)


parametrization:
Requestor: Tariq.Morales
Requestee: Paloma.Gupta
RequestOrReplied: Replied

Must-fix: None
Tech-debt: Inconsistent-state hardening (non-blocking, theoretical) — see below.

Approve. This is the correct one-layer-down sibling of #100 and it resolves the zero-PR scoping cleanly.

Correctness: `_wave_ordinal_for_wave` mirrors `_phase_for_wave` exactly — same read/decode/None-on-error posture, same `val is not None` guard so a legitimate ordinal of `0` would survive. `_integration_base` substitutes `wave_ordinal` for the `{wave}` token only when supplied and otherwise falls back to the global wave id, which keeps generic `deployments/wave-{wave}` projects rendering unchanged. I confirmed the live coupling holds end to end: `lifecycle.py:428` stamps `wave_<id>_phase_ordinal` in the same dict as `wave_<id>_phase`, and `.claude/state.json` carries `wave_2_phase_ordinal: 1` — so scoring global wave 2 now resolves `deployments/phase4/wave-1` (phase=4, ordinal=1) with no override, exactly as the body claims. The re-score prerequisite is satisfied.

Test coverage (QA lens): strong and correctly targeted. The `_wave_ordinal_for_wave` reader has the full quartet (happy path, missing key, no-path, unreadable) matching the `_phase_for_wave` suite; `test_integration_base_uses_phase_local_ordinal` pins the exact bug (global 2 -> wave-1, not wave-2); `test_integration_base_ordinal_falls_back_to_wave_id` locks the generic-project fallback (both implicit and explicit `wave_ordinal=None`); and `test_base_resolves_phase4_wave1_from_global_wave_2` is a genuine end-to-end regression guard resolving BOTH tokens from a state file. The fixtures use the real Phase 4 shape rather than toy values, which I like. No coverage gaps that block.

Dual-deploy (#116): compliant. Canonical `framework/assets/lib/trust_signals.py` edited; I verified there is no live `.claude/lib/trust_signals.py` installed copy in this repo (only ephemeral worktree checkouts), so no source/runtime drift today — the deferred reinstall note is accurate. CI green across all matrices. Diff is disjoint from #119 (touches `_wave_ordinal_for_wave`/`_integration_base`/`merged_prs` + a separate test section), consistent with the sequencing note.

Tech-debt (non-blocking): if a state file ever carries `wave_<id>_phase` but not `wave_<id>_phase_ordinal`, `_integration_base` falls back to the global wave id and re-renders the old buggy `deployments/phase4/wave-2`. Today that can't happen because the allocator stamps both keys atomically, so this is purely defensive — a follow-up could tighten it (e.g. derive the ordinal from phase when only phase is present) if the state schema ever splits those writes. Not merge-blocking.

PR #125, opened 2026-07-05
fix(ontology-gen): exclude installed framework runtime in consumer repos (#74)

parametrization:
Closes #74.

## Problem
In a fresh target repo the generated structural index was dominated by the framework's
own installed runtime. A bootstrap install lays ~35 internals under `.claude/hooks/` +
`.claude/lib/`, and `discover()` only skipped `.claude/worktrees/` — so a scratch install
with a single `app.py` produced 36 indexed files, 35 of them framework internals.

## Fix
`ontology_gen.generate.discover()` now excludes the whole installed `.claude/` tree in
**consumer** repos. The framework **source** checkout is detected by the canonical install
dirs it ships (`framework/assets` + `framework/install`); there `.claude/` is a live dogfood
copy and stays indexed (current behaviour preserved). Zero-config — no new config keys.

## Tests (`framework/tests/test_ontology_gen.py`)
- `test_generate_skips_installed_runtime_in_consumer_repo` — consumer repo with
  `.claude/hooks` + `.claude/lib` indexes only `app.py`.
- `test_generate_indexes_dogfood_claude_in_framework_source_repo` — source-marked repo
  still indexes its `.claude/` dogfood copy.

`pytest framework/tests/test_ontology_gen.py` → 18 passed. `ruff check` clean.

### DIFF
--- a/framework/assets/lib/ontology_gen/generate.py
+++ b/framework/assets/lib/ontology_gen/generate.py
@@ -52,10 +52,30 @@ _IGNORE_DIRS = frozenset(
 )
 # Always skipped even under git discovery (scratch checkouts of other branches).
 _ALWAYS_SKIP_PREFIXES = (".claude/worktrees/",)
+# The framework's OWN installed runtime in a consumer repo (issue #74). A bootstrap install
+# lays ~35 framework internals under .claude/hooks/ + .claude/lib/; indexing them floods a
+# target repo's structural index with framework noise instead of the user's code. Skipped in
+# consumer repos only — the framework SOURCE repo's .claude/ is a live dogfood copy worth
+# indexing, so it keeps the current behaviour (see _is_framework_source_repo).
+_INSTALLED_RUNTIME_PREFIX = ".claude/"
 # Vendored deps / build output — skipped in BOTH discovery modes, even when committed.
 # git ls-files trusts tracked files, so repos that commit node_modules/dist/coverage would
 # otherwise flood the index with thousands of vendored/generated files.
 _ALWAYS_SKIP_DIRS = frozenset({"node_modules", "dist", "build", "coverage"})
+
+
+def _is_framework_source_repo(repo_root: Path) -> bool:
+    """True when ``repo_root`` is the framework SOURCE checkout, not a consumer install.
+
+    The source repo ships the canonical install dirs (``framework/assets`` +
+    ``framework/install``); a repo that merely *installed* the framework has a populated
+    ``.claude/`` but none of that source tree. This gate decides whether the installed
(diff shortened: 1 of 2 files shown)


parametrization:
Requestor: Tariq.Morales
Requestee: Ibrahim.El-Amin
RequestOrReplied: Replied

Must-fix: None
Tech-debt: Source-detection hardening + two missing regression guards — see below.

Approve. The consumer-vs-source split is the right shape and both directions are tested.

Correctness: `discover()` conditionally extends `skip_prefixes` with `.claude/` only when `_is_framework_source_repo()` is false, so a consumer install drops its whole installed `.claude/` runtime while the source checkout keeps indexing its dogfood copy. I confirmed the gate behaves correctly for THIS repo: `framework/assets` and `framework/install` both exist, so `_is_framework_source_repo` returns true and this repo's own ontology output is unchanged (no dogfood regression). Adding `.claude/` on top of the existing `_ALWAYS_SKIP_PREFIXES` (`.claude/worktrees/`) is redundant-but-harmless in the consumer path since the worktrees prefix is a subset. The `startswith(tuple)` call handles the combined tuple correctly.

Test coverage (QA lens): both critical directions are covered — `test_generate_skips_installed_runtime_in_consumer_repo` proves a consumer repo indexes only `app.py` (asserts count AND absence of `.claude/hooks`/`.claude/lib` in llms.txt), and `test_generate_indexes_dogfood_claude_in_framework_source_repo` proves the source-marked repo still indexes its `.claude/` dogfood copy. Good use of real install-dir markers rather than mocks. This clears the bar to merge.

Dual-deploy (#116): compliant. Canonical `framework/assets/lib/ontology_gen/generate.py` edited; no live `.claude/lib/ontology_gen/generate.py` installed copy exists in this repo (only ephemeral worktree checkouts), so no runtime drift today. CI green across all matrices; diff is self-contained.

Tech-debt (non-blocking follow-ups):
1. Detection robustness: `_is_framework_source_repo` keys on two generic directory NAMES (`framework/assets` + `framework/install`). A consumer repo that coincidentally vendors those paths would be misclassified as source and re-index its installed runtime. Unlikely, but a more specific sentinel (e.g. a shipped marker file like `framework/install/bootstrap.py`, or a `framework/VERSION`) would harden it. Consider for a later pass.
2. Missing partial-marker test: the `and` in the detector is untested — there is no case where exactly one of the two dirs is present (should classify as consumer). Add one to pin the AND semantics against future edits.
3. Missing regression guard: no test asserts `.claude/worktrees/` stays skipped in a SOURCE repo (i.e. that the source path still applies `_ALWAYS_SKIP_PREFIXES`). Cheap to add and guards the redundancy assumption.

None of these block; the shipped behavior is correct and covered.

PR #127, opened 2026-07-05
fix(node): refresh stale framework bundle on prepack instead of skip-if-exists (#82)

parametrization:
## Issue #82 — `copy-shared.js` skip-if-exists can publish a stale `node/framework/` bundle

`node/scripts/copy-shared.js` (the `npm run prepack` bundler) skipped any destination directory that already existed. `templates/`/`presets/`/`skills/` are stable, but `framework/` — the config-driven runtime — changes **every wave**. A developer who ran `npm pack` once locally kept re-publishing that first snapshot on every subsequent local pack, shipping an outdated runtime. CI publishes from a fresh checkout (OIDC), so released tarballs were always safe; this was a local footgun only, but a real one.

### Fix
Replace skip-if-exists with **rm+recopy for every non-symlink destination**, so each prepack ships the current tree:
- A destination that is a **symlink** (e.g. `node/framework -> ../framework` for live dev) is detected via `lstat` and left untouched — the escape hatch the original skip was likely protecting. Never clobbered, never copied onto itself.
- A destination that is a **real directory** (a leftover snapshot from an earlier pack) is removed and recopied, guaranteeing freshness and clearing files dropped upstream.
- `tests/` / `__pycache__` / `.pyc` exclusions unchanged.

### Verify
- New `node/tests/copy-shared.test.ts` (5 cases) copies the **real** script into a throwaway package layout and runs it with `node`, exercising `__dirname/../..` REPO_ROOT resolution for real: first-pack copy, **stale-bundle refresh (v1→v2)**, leftover-file drop on re-sync, **symlink preservation**, and dev-subtree exclusion.
- Full node suite `vitest run` = **132 passing**.

### TL cross-issue note (#82 ↔ #116)
#116 (Ibrahim, reinstall) may re-run the node install path. This fix makes the local pack path deterministic and self-refreshing, so a reinstall that re-packs will always pick up the current `framework/` runtime rather than a cached snapshot — no interaction needed beyond "reinstall benefits from this being fixed." No shared files touched with #116.

Files: `node/scripts/copy-shared.js`, `node/tests/copy-shared.test.ts`.

Base: `deployments/phase4/wave-2`. Do not merge without owner approval.

### DIFF
--- a/node/scripts/copy-shared.js
+++ b/node/scripts/copy-shared.js
@@ -17,7 +17,7 @@
  * __pycache__/ or tool cache directories.
  */
 
-import { cpSync, existsSync, mkdirSync } from "node:fs";
+import { cpSync, existsSync, lstatSync, mkdirSync, rmSync } from "node:fs";
 import { resolve, dirname, basename } from "node:path";
 import { fileURLToPath } from "node:url";
 
@@ -48,9 +48,20 @@ for (const dir of DIRS) {
     continue;
   }
 
-  if (existsSync(dest)) {
-    // Already copied or symlinked — skip
+  // Symlink escape hatch: a developer may symlink e.g. node/framework -> ../framework
+  // for a live checkout. Never clobber that — leave the link in place.
+  const link = lstatSync(dest, { throwIfNoEntry: false });
+  if (link && link.isSymbolicLink()) {
+    console.log(`Skipped ${dest} (symlink — live checkout)`);
     continue;
+  }
+
+  // A real directory left over from an earlier local pack is a STALE snapshot:
+  // these bundled trees (framework/ in particular) change every wave, and a
+  // skip-if-exists would silently re-publish the old copy (issue #82). Remove and
+  // recopy so every prepack ships the current tree.
+  if (existsSync(dest)) {
+    rmSync(dest, { recursive: true, force: true });
   }
 
   mkdirSync(dest, { recursive: true });
--- a/node/tests/copy-shared.test.ts
+++ b/node/tests/copy-shared.test.ts
@@ -1,0 +1,119 @@
+/**
+ * Tests for scripts/copy-shared.js — the prepack bundler that copies the shared
+ * data trees (templates/, presets/, skills/, framework/) into the node package so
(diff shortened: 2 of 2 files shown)


parametrization:
Requestor: Paloma.Gupta
Requestee: Nia.Rossi
RequestOrReplied: Replied

Must-fix: None
Tech-debt: None

Charter-format review of #82 (copy-shared.js stale-bundle refresh). Approving — ships as-is.

**Symlink detection is correct.** `lstatSync(dest, { throwIfNoEntry: false })` returns `undefined` on a missing dest (no throw), and `isSymbolicLink()` inspects the link itself because `lstat` does not dereference — so the `node/framework -> ../framework` live-checkout escape hatch is left untouched, never clobbered, and never copied onto itself. Ordering is right: symlink branch first, then `existsSync` gate on the rm+recopy (a missing dest short-circuits `existsSync` to false → falls through to the clean mkdir+copy), then mkdir+cpSync. A broken symlink is also preserved, which is the correct escape-hatch behavior.

**Node 18/20/22 all pass.** Every fs API used predates 18: `rmSync` (14.14), `lstatSync`'s `throwIfNoEntry` option (15.3 / 14.17 backport). CI's `node` job matrix is `[18, 20, 22]` running `npm run build` + `npm test`, so the matrix exercises this path on all three. The test writing `package.json {"type":"module"}` into the throwaway layout to force ESM under Node 18 (no syntax auto-detect) is a nice touch that keeps the black-box run honest across the matrix.

**Test coverage is thorough and real.** Copying the actual script into a fake repo and running it through `node` exercises the `__dirname/../..` REPO_ROOT resolution for real, and the five cases cover the load-bearing behaviors: first-pack copy, stale v1→v2 refresh (the #82 regression), leftover-file drop on re-sync, symlink preservation (asserts both the read-through and that the source wasn't clobbered), and the tests/__pycache__ exclusion.

Dual-deploy: this is the bundler script, not a `.claude/` asset, so no mirror surface — no #116 interaction beyond the correctly-noted "reinstall benefits from a deterministic pack." Coordination note with #116 reads accurately.

PR #128, opened 2026-07-05
test(install): cover meta-and-children ontology generation at install time (#75)

parametrization:
## Issue
Closes #75 — no test coverage for meta-and-children ontology generation at install time.

## Context
`generate_structural()` (bootstrap.py) threads `project.model` through to `ontology_gen.refresh.refresh()`, which runs the `meta-and-children` cross-repo aggregation. Coverage existed only for single-repo installs (`test_bootstrap_smoke.py`), so issue #66's AC "works for meta installs" was unverified. The pass-through was correct by inspection; this PR pins it end-to-end.

## What was added
`framework/tests/test_meta_child_install.py` → new `TestMetaOntologyInstall` (subprocess bootstrap into a real meta root):
- **`test_meta_install_emits_cross_repo_aggregate`** — meta root + two seeded child git-repos + parent source; asserts `project.model == meta-and-children` in the written config, the `cross-repo N nodes across M repo(s)` report line (3 repos: parent + api + web), the written `ontology/structural/cross-repo-graph.json`, per-repo namespacing (children `api/`,`web/` + parent keyed by `root.name`), and that the aggregate node count matches the reported count.
- **`test_meta_install_aggregate_is_idempotent`** — re-run reports `structural index: fresh` and leaves the aggregate byte-identical (deterministic + staleness-guarded).
- **`test_meta_install_missing_child_index_degrades_gracefully`** — a child git-repo with no per-repo index is skipped (2 repos, not 3) and the install still succeeds — pins the aggregator's graceful-degradation contract at install time.

Helper `_seed_child_index` pre-generates a child's per-repo index (children do not self-generate during a meta install; the parent aggregates whatever indices are present on disk — mirrors a committed child index reaching the roll-up).

## Verification
Test-only; **no runtime change**. `ruff check framework/` clean; full framework suite green (334 passed).

## Coordination
- **#65 (meta install path):** builds directly on that path; no code change to it, only coverage.
- Dual-deploy: test-only, exercises the canonical `framework/**` install path; no `.claude/` mirror or reinstall implication.

### DIFF
--- a/framework/tests/test_meta_child_install.py
+++ b/framework/tests/test_meta_child_install.py
@@ -10,6 +10,7 @@ table, and the interactive children multi-select (input stubbed).
 from __future__ import annotations
 
 import json
+import re
 import subprocess
 import sys
 from pathlib import Path
@@ -19,10 +20,23 @@ import pytest
 _FRAMEWORK_ROOT = Path(__file__).resolve().parent.parent
 _BOOTSTRAP = _FRAMEWORK_ROOT / "install" / "bootstrap.py"
 sys.path.insert(0, str(_FRAMEWORK_ROOT / "install"))
+sys.path.insert(0, str(_FRAMEWORK_ROOT / "assets" / "lib"))
 
 import bootstrap  # noqa: E402
 import child_install  # noqa: E402
 import roster_gen  # noqa: E402
+from ontology_gen.generate import generate as _generate_index  # noqa: E402
+
+
+def _seed_child_index(repo_root: Path, name: str) -> dict[str, int]:
+    """Pre-generate a child's per-repo structural index (simulating a COMMITTED
+    child index). During a meta install the parent aggregates whatever per-repo
+    indices are present on disk; children do not self-generate at install time,
+    so this is how a real child index reaches the roll-up.
+    """
+    out = repo_root / "ontology" / "structural"
+    out.mkdir(parents=True, exist_ok=True)
+    return _generate_index(repo_root, out, name)
 
 
 def _run(*argv: str) -> subprocess.CompletedProcess:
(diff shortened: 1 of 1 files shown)


parametrization:
Requestor: Ibrahim.El-Amin
Requestee: Tariq.Morales
RequestOrReplied: Replied

Must-fix: None
Tech-debt: None

Charter-format review. Approving. I authored the #74 ontology-gen fix, so I checked every cross-repo assertion against the actual generate/aggregate/refresh/bootstrap behavior — they are all sound:

- **model normalization.** `install-config model: meta` → runtime `meta-and-children` via `RUNTIME_MODEL_MAP` (`install_config.py`); the assertion `project.model == "meta-and-children"` is correct, and only that runtime value makes `refresh()` take the aggregate branch.
- **report line.** `bootstrap.py` emits `"; cross-repo {nodes} nodes across {repos} repo(s)"` — the `_CROSS_REPO_RE` matches it exactly.
- **repos count.** `refresh()` sets `aggregate.repos = sum(1 for s in statuses if s.present)`, i.e. present indices only. So the degradation case (parent + api present, web index absent) correctly reports **2**, and the graceful-skip contract (`_load_namespaced` → None → `present=False`, not added to the graph) is what makes `"web" not in namespaces` hold without aborting the install. That is the exact contract I most wanted pinned.
- **namespacing / parent keying.** `_load_namespaced` prefixes every `path` with `<name>/`; `discover_repos` keys the parent by `root.name` (subdir `.`) alongside child git-repos. So `{"api","web"} <= namespaces` and `"meta" in namespaces` are both right, and `len(graph["nodes"]) == nodes` follows from the reported count coming off the assembled graph.
- **idempotency.** On re-run the parent index is not stale → `refresh` returns `{"regenerated": False, "status": "fresh"}` early, so bootstrap prints `structural index:  fresh (already current)` (substring match holds) and the aggregate is left byte-identical. Correct — and it also implicitly documents that the roll-up rides on the per-repo staleness guard.

`_seed_child_index` mirroring a committed child index is the right model — children don't self-generate during a meta install; the parent aggregates whatever per-repo indices are on disk. Test-only, no runtime change, exercises the canonical `framework/**` install path end-to-end. This is the coverage #66/#75 were missing.

PR #154, opened 2026-07-06
harness: real-repo provisioner (clone-at-pinned-SHA) + wire --include-real for B10/B11 (#153)

parametrization:
## What

Replaces the B10/B11 `_prov_real_stub` (`framework/harness/buckets.py`, which unconditionally
raised `NotImplementedError`) with a real **clone-at-pinned-SHA** provisioner and wires
`--include-real` so B10 (`meta-real-world`) and B11 (`standalone-real-world`) are actually
provisioned + run instead of hitting the stub.

New module: `framework/harness/real_provision.py`.

## Mechanism — read-only w.r.t. the live source (the load-bearing safety property)

Other sessions are live-committing in the real source trees, so provisioning never touches them:

- **`git clone --no-local <source> <scratch>`** forces a full object copy — never a hardlink into
  the source `.git` (the plain local-clone default), never a worktree/copy of the live tree.
  Working-tree dirt in the source is therefore never carried into the clone; a **locally-dirty
  source still clones clean at the pin**.
- **Pin resolution** — an explicit SHA verbatim, else resolved from the source's
  `refs/heads/main` via `git ls-remote` **at run time** (never the locally checked-out branch,
  since several sources sit on in-progress feature branches).
- **Source-unchanged invariant** — each local source's `{HEAD, git status --porcelain}` is
  fingerprinted before and after every provision and asserted **byte-identical**; a mismatch
  raises `SourceMutatedError` (isolated to one errored cell, not a skip).

## Source/pin registry is DATA (not hardcoded in flow)

`RealFixtureSpec` / `RealChildSpec` dataclasses + a `DEFAULT_REAL_FIXTURES` map, overridable via a
`--real-config` sidecar JSON or (in tests) `opts['real_fixtures']`. B10 meta clones the parent +
its gitignored sibling children into the same layout the synthetic `_prov_meta` produces, so child
wiring resolves identically. **#101/#109 supply real source paths / pins / noorinalabs children via
`--real-config`** — no real paths are load-bearing here.

Registry shape (for #101/#109):
```json
{"B11": {"source": "<path-or-remote>", "pin": "<sha|null>", "ref": "refs/heads/main"},
 "B10": {"source": "<parent>", "model": "meta-and-children",
         "children": [{"path": "child-dir", "source": "<child>", "pin": null, "flavor": "product"}]}}
```

## Degrade + teardown

- A genuinely unresolvable fixture (missing path / unreachable remote) raises `MissingFixtureError`
  (a `NotImplementedError` subclass) so the runner's **pre-existing stub-skip path** fires — CI and
  machines without the local checkouts stay green; it never fires by default on a dev box that has
  the fixtures.
- **Zero-residue teardown** reuses the existing drop-the-copy `rmtree` + the symmetric-diff proof
  (`teardown_residue_zero` + `teardown_proof=True` on both real legs) — a real bucket leaves an
  empty diff.

## Hermetic test approach (no real fixtures, no network)

`framework/tests/test_real_provisioner.py` builds a throwaway local git repo in `tmp_path` and uses
it as the clone source. It proves: clone-at-pin (pinned commit, not the tip), dirty-source-clones-
clean, the **source-unchanged invariant** (feature-branch + dirty source is byte-identical after a
provision AND after a full matrix run), `--include-real` executes the real B10/B11 path all-green
(not the stub), explicit-pin and `ls-remote`-resolved pin, missing-fixture → skip (no crash), zero
scratch residue, and the meta parent+children clone layout. The real multi-GB noorinalabs/botfarm
runs are **#101 / #109**, deliberately not exercised here.

## Scope note

`framework/harness/**` is dev/test tooling, **not** an installed asset under `framework/assets/` or
`framework/install/` — **#116 dual-deploy does NOT apply**; there is no canonical-copy diff to look
for.

## Quality

- `python3 -m pytest framework/tests/ -q` → **456 passed** (was 442; +14 hermetic tests, 1 existing
  stub test rewritten in place).
- End-to-end smoke: `python3 -m framework.harness --buckets B11 --include-real --real-config <json>`
  against a synthetic repo → B11 green (12/12 graded), source `git status --porcelain` empty and
  still on `main` afterward.

Closes #153

**Reviewers: Nia Rossi** (design/architecture) + **Tariq Morales** (QA/teardown) — named here
because simulated teammates can't be added as GitHub reviewers.

### DIFF
--- a/framework/harness/__main__.py
+++ b/framework/harness/__main__.py
@@ -33,6 +33,9 @@ def _parse_args(argv: list[str]) -> argparse.Namespace:
                     help="comma-separated installers to run (bootstrap,cli)")
     ap.add_argument("--include-real", action="store_true",
                     help="opt in to the [real] B10/B11 buckets (owner-gated; OFF by default)")
+    ap.add_argument("--real-config", type=Path, default=None,
+                    help="sidecar JSON overriding the B10/B11 real-fixture source/pin registry "
+                         "(see framework.harness.real_provision.RealFixtureSpec)")
     ap.add_argument("--no-dogfood", action="store_true", help="skip the B12 reinstall --check leg")
     ap.add_argument("--scale", type=int, default=None, help="B9 file count (default: 300)")
     ap.add_argument("--out", type=Path, default=_DEFAULT_RUNS_DIR,
@@ -64,6 +67,8 @@ def _opts(args: argparse.Namespace) -> dict:
     }
     if args.scale is not None:
         opts["scale"] = args.scale
+    if args.real_config is not None:
+        opts["real_config"] = str(args.real_config)
     return opts
 
 
--- a/framework/harness/buckets.py
+++ b/framework/harness/buckets.py
@@ -20,6 +20,8 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Callable
 
+from .real_provision import make_real_provisioner
+
 _FRAMEWORK_ROOT = Path(__file__).resolve().parent.parent
 _BOOTSTRAP = _FRAMEWORK_ROOT / "install" / "bootstrap.py"
 
@@ -202,12 +204,9 @@ def _prov_large(wd: Path, opts: dict) -> dict:
     return {"extra": {"expect_source_path": "gen/f00000.py"}, "n_files": n}
 
 
-def _prov_real_stub(wd: Path, opts: dict) -> dict:
-    raise NotImplementedError(
-        "B10/B11 real-world provisioning is flag-gated and NOT wired in this wave. When enabled "
-        "it must CLONE the remote default branch at a pinned SHA into this scratch dir — never "
-        "copy/worktree the live working repo (other sessions operate there). See owner-decision 1."
-    )
+# B10/B11 real-world provisioning (clone-at-pinned-SHA into scratch, read-only w.r.t. the live
+# source) lives in ``real_provision`` and is bound per-bucket by ``make_real_provisioner`` — the
+# source/pin registry is DATA there (#153), overridable via ``--real-config`` for #101/#109.
 
 
 # --------------------------------------------------------------------- the matrix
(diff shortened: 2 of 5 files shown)


parametrization:
Requestor: Nia.Rossi
Requestee: Paloma.Gupta
RequestOrReplied: Replied

**Review: architecture/design — sound mechanism, clean skip-path reuse, approve with tech-debt**

Checked the source-unchanged invariant, pin resolution, registry-as-data shape, `--include-real`
wiring, and scope discipline against `real_provision.py` + the `buckets.py`/`__main__.py` wiring
and `test_real_provisioner.py`.

- **Invariant**: `git clone --no-local` only ever reads the source and detaches `HEAD` on `dest`
  (never `source`) — the actual read-only guarantee is a property of the git commands chosen, not
  just the fingerprint. `resolve_pin()` correctly reads `refs/heads/main` via `ls-remote` rather
  than the checked-out branch, proven directly by `test_resolve_pin_reads_remote_main_not_local_checkout`
  (dirty feature-branch source, remote ref still wins). `SourceMutatedError` correctly does NOT
  subclass `NotImplementedError`, so a real mutation surfaces as a hard cell failure, not a silent
  skip — right call.
- **Registry shape**: `RealFixtureSpec`/`RealChildSpec` mirror `_prov_meta`'s parent+children
  layout closely enough that #101/#109 should be able to point `--real-config` at real sources
  without touching `real_provision.py` itself.
- **`--include-real` wiring**: `MissingFixtureError(NotImplementedError)` cleanly rides the
  pre-existing stub-skip path in `runner.run_cell` — no new control flow in the runner.
- **Scope**: confined to `framework/harness/`, hermetic tests (throwaway local repos in
  `tmp_path`, no network), `#116` correctly N/A.

No blocking holes in the invariant or the registry API. Findings below are all non-blocking.

Must-fix: None
Tech-debt:
1. `buckets.py`'s module docstring (top of file, "Owner decisions") still says B10/B11
   provisioning "is a guarded stub" — now stale/self-contradicting against the accurate comment
   a few lines below it and against `real_provision.py` itself. Update the wording.
2. `provision_real()`'s `before`/`after` fingerprint compare is skipped entirely if `clone_at`
   raises partway through a multi-clone provision (e.g. a B10 child source unreachable after the
   parent clone succeeded) — the exception propagates before `after` is ever computed. Not an
   actual safety hole (`clone_at` never writes to `source` regardless), but it weakens the
   defense-in-depth the docstring promises ("every provision asserts..."). Worth wrapping in
   try/finally so the invariant is still checked (or explicitly explained) on the error path.
3. `DEFAULT_REAL_FIXTURES` hardcodes this dev machine's absolute home-directory paths
   (`/home/parameterization/code/...`) in checked-in source. Harmless today (gated off by
   default, degrades to `MissingFixtureError`→skip anywhere else), but worth a comment noting
   these are dev-machine placeholders, not portable defaults.
4. `real_registry()`'s `--real-config` / `real_fixtures` overrides fully replace a bucket's spec
   rather than merging fields — e.g. overriding just `pin` for B10 silently drops `children` back
   to `()` unless the sidecar repeats the full children list. Worth a one-line docstring note so
   #101 doesn't get bitten by a partial override.
5. `DEFAULT_REAL_FIXTURES["B10"]` ships with `children=()` — a `--include-real` run on B10 without
   a `--real-config` override produces a degenerate meta-with-zero-children install (harmless, no
   crash, just semantically vacuous until #101 populates real children).
6. `RealChildSpec.path`/`clone_at` assume a single-level child directory (`wd / child.path`
   without `mkdir(parents=True)` first) — a nested child path (e.g. `"libs/api"`) would fail
   clone with "could not create leading directories" and degrade to a skip rather than a clean
   error. Not exercised by any current fixture; flag for #101 if nested child layouts show up.

parametrization:
Requestor: Tariq.Morales
Requestee: Paloma.Gupta
RequestOrReplied: Replied

**Review: Must-fix resolved — negative-path guard for the read-only invariant now covered; clears QA**

_Amended in place after re-verifying commit `dab0682` (independent scratch clone of
`P.Gupta/0153-real-repo-provisioner`, main worktree untouched). Original verdict was
ChangesRequested with one Must-fix; that item is now fixed._

The Must-fix is resolved by `test_source_mutation_mid_provision_raises_source_mutated`
(`framework/tests/test_real_provisioner.py`): it monkeypatch-wraps `clone_at` to write an
`intruder.txt` into the live source BETWEEN `provision_real`'s before/after fingerprints — a
faithful stand-in for a concurrent session dirtying the tree mid-provision — and asserts
`SourceMutatedError` raises, that the error names the mutated source path (proving the
after-fingerprint mismatch is the trigger, not some other failure), and that the source genuinely
changed. I confirmed it is a real regression guard, not a tautology: I neutered the invariant
(`if not unchanged:` → `if False:`) and the test flipped red with `Failed: DID NOT RAISE
SourceMutatedError`; restored, it passes. `real_provision.py` has zero net diff vs the reviewed
commit (the only production change is a buckets.py top-of-file docstring that no longer calls B10/B11
"a guarded stub" — Nia's tech-debt #1, folded in). Full suite 456→457 passed, `ruff check
framework/` clean.

Must-fix: None
Tech-debt:
1. `clone_at`'s checkout-of-a-bad-SHA branch (clone succeeds, then `git checkout --detach <sha>`
   fails because an explicit `--real-config` pin doesn't exist) isn't exercised — only
   `resolve_pin`'s ls-remote-failure path is. Same for a `git clone` failure not caused by an
   unresolvable ls-remote (e.g. a `source` path that exists but isn't a git repo). Non-blocking.
2. The CLI-level `--real-config` argparse → `_opts()` wiring in `__main__.py` has no end-to-end
   test; only the underlying `real_registry()` sidecar-merge is unit-tested. Low risk (few lines,
   simple pass-through) but worth a smoke test given #101/#109 will drive it for real. Non-blocking.

PR #160, opened 2026-07-06
fix(install): durability hardening from #101/#109 real runs (#149)

parametrization:
## Durability + fidelity hardening from the #101/#109 real-repo runs (#149)

Triaged the durability findings from the noorinalabs (#101) and botfarm_inc (#109) real
runs; landed the clear wins with tests and split the two redesign-sized items out as
tracked deferrals (comment on #149).

### Fixes (one test per fix)

- **`atomic_io.atomic_write_text` parent-dir fsync gap** (finding 2) — after `os.replace`
  we now fsync the target's parent directory so the rename's directory entry is durable
  across a power loss (fsyncing the temp file persisted the bytes but not the name). New
  `_fsync_dir` helper is best-effort / fail-open: platforms that can't open a directory fd
  (Windows) or fsync one are a silent no-op and never crash the install.
- **`repo_space.archive_assets` non-atomic move-then-manifest** (finding 3) — the restore
  manifest is now written **before** the first `shutil.move`, so a crash mid-archive leaves
  a fully manifest-recoverable archive instead of a stranded, unrestorable half. `restore`
  moves back whatever reached the archive and skips members still untouched at the repo root.
- **`no_backup_litter` counted the source repo's own checked-in `.bak`** (from #101) — the
  harness metric now diffs `.bak*` against the pre-install snapshot (`ctx.before`), so a
  fixture's own `errors.jsonl.bak.*` no longer reads as installer litter; only installer-
  created backups count.
- **`.claude-backups/` not gitignored** (finding 5) — added to the repo-root `.gitignore`
  so archive/restore leaves no untracked noise.
- **Restore out-of-scope residue** (finding 4) — documented that restore's managed scope is
  exactly `.claude/` + `CLAUDE.md`; `.git/hooks/pre-push` and the emptied `.claude-backups/`
  container are intentionally left (pruning to "globally pristine" is a separate concern). A
  test asserts the boundary is exact.

### Deferred (redesign-sized; tracked in a comment on #149, no parallel issue)

1. In-place `settings.json` merge not reversible by manifest-driven teardown (finding 1) —
   changing the amend contract to drop a `.bak`/restore entry for any pre-existing file it
   mutates.
2. B10 "pre-existing-framework baseline mode" metric-modeling — grading install-over-existing
   correctly across the metric set.

(Original Phase-5 issue-body items 2 & 3 — content-based foreign-asset detection and
child/meta golden-manifest snapshots — are out of this PR's scope and stay open on #149.)

### Verification

- `python3 -m pytest framework/tests/ -q` → **465 passed** (base 460 on `c1fe0b2`, +5 new).
- **#116**: all edits are installer (`framework/install`) / dev tooling (`framework/harness`)
  / tests / repo-root `.gitignore` — no `.claude/**` runtime asset touched, so reinstall-
  parity does not apply.

Addresses #149

Reviewers: Paloma Gupta + Tariq Morales

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,9 @@ node/*.tgz
 .claude/settings.local.json
 # Agent worktree checkouts (isolation dirs, never committed)
 .claude/worktrees/
+# Consent-gated archive-then-fresh backups (repo_space.py archive/restore, #108/#149);
+# a restore/archive must not leave untracked noise in the working tree.
+.claude-backups/
 
 # ---- Editor / OS ----
 .DS_Store
--- a/framework/harness/metrics.py
+++ b/framework/harness/metrics.py
@@ -510,7 +510,18 @@ def m_teardown_residue_zero(ctx: CellContext) -> Measurement:
 
 def m_no_backup_litter(ctx: CellContext) -> Measurement:
     # A happy-path fresh install (no conflict) must create zero .bak* files.
-    baks = [p.name for p in ctx.workdir.rglob("*.bak*") if ".git" not in p.parts]
+    # Scope detection to files the INSTALLER created: a .bak already present in the
+    # pre-install snapshot (ctx.before) is the source repo's own checked-in backup — e.g.
+    # noorinalabs's `errors.jsonl.bak.*` from #101 — not installer litter, so exclude it by
+    # diffing against the pre-install snapshot rather than counting every `.bak*` in the tree.
+    preexisting = set(ctx.before)
+    baks = []
+    for p in ctx.workdir.rglob("*.bak*"):
+        if ".git" in p.parts or not p.is_file():
+            continue
+        if p.relative_to(ctx.workdir).as_posix() in preexisting:
+            continue  # checked into the fixture before install — the repo's litter, not ours
+        baks.append(p.name)
     ok = not baks
     return Measurement(len(baks), ok, {"bak_files": 0}, {"bak_count": len(baks), "baks": baks[:20]})
 
--- a/framework/install/atomic_io.py
+++ b/framework/install/atomic_io.py
@@ -11,16 +11,22 @@ a backup, but the write itself should never be able to strand a half-file.
 the SAME directory, flushes + fsyncs it, then ``os.replace``s it onto the target.
 ``os.replace`` is an atomic rename on POSIX (and same-volume on Windows), so any
 concurrent reader — and any post-crash observer — sees either the old file intact
-or the complete new file, never a truncated in-between. Stdlib-only.
+or the complete new file, never a truncated in-between. After the replace we also
+fsync the *parent directory* so the rename's new directory entry is durable across a
+power loss — fsyncing the temp file persists its bytes, but not the entry that names
+them under the target, so a crash right after the rename could otherwise lose the
+file on some filesystems. The dir-fsync is best-effort / fail-open: platforms that
+cannot open a directory for fsync (notably Windows) skip it silently and never crash
+the install. Stdlib-only.
 
 Public API
 ==========
 ``atomic_write_text(path, text, *, encoding="utf-8") -> None``
     Create ``path``'s parent if needed, write ``text`` to a temp file in the same
     directory (so the final rename stays on one filesystem and is therefore
-    atomic), then ``os.replace`` it into place. The temp file is removed on any
-    error before the replace, so a failed write leaves neither a stray temp nor a
-    damaged target.
+    atomic), then ``os.replace`` it into place and fsync the parent directory so the
+    rename is durable. The temp file is removed on any error before the replace, so a
+    failed write leaves neither a stray temp nor a damaged target.
 """
 
 from __future__ import annotations
@@ -28,6 +34,28 @@ from __future__ import annotations
 import os
 import tempfile
 from pathlib import Path
+
+
+def _fsync_dir(directory: Path) -> None:
+    """Best-effort fsync of ``directory`` so a rename into it survives a power loss.
+
+    ``os.replace`` puts the new name in the directory, but that directory entry may live
+    only in the OS cache until the *directory itself* is fsynced; a crash in that window can
+    lose the rename even though the file's bytes were fsynced. Opening the dir and fsyncing
+    its fd flushes the entry. Fail-open by contract: platforms/filesystems that cannot open a
+    directory for reading (notably Windows) or cannot fsync one are a silent no-op — hard
+    durability is best-effort and must never crash the install.
+    """
+    try:
+        dir_fd = os.open(str(directory), os.O_RDONLY)
+    except OSError:
+        return  # e.g. Windows: cannot open a directory as a file descriptor
+    try:
+        os.fsync(dir_fd)
+    except OSError:
+        pass  # some filesystems reject fsync on a directory fd — durability stays best-effort
+    finally:
+        os.close(dir_fd)
 
 
 def atomic_write_text(path: Path | str, text: str, *, encoding: str = "utf-8") -> None:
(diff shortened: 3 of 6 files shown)


parametrization:
Requestor: Paloma.Gupta
Requestee: Ibrahim.El-Amin
RequestOrReplied: Replied

Reviewed the diff directly plus the checked-out branch (`465 passed`, `ruff check` clean —
reproduced both locally). Traced each fix's crash-consistency semantics rather than reading
the tests as sufficient proof on their own.

**1. `_fsync_dir(path.parent)` after `os.replace` — correct discipline, verified ordering.**
`atomic_write_text` is: write temp → `fh.flush()` + `os.fsync(fh.fileno())` → `os.replace(tmp,
path)` → (only on the success path, since `_fsync_dir` sits *outside* the `try/except
BaseException` block) `_fsync_dir(path.parent)`. That's the right sequence — temp-file fsync
makes the bytes durable, dir fsync makes the *name* durable; without the latter a crash right
after the rename could resurrect the old file or an empty dir entry on some filesystems even
though the bytes are safely on disk. Confirmed the temp-fsync line is untouched (still
present, still before the replace). `_fsync_dir` only catches `OSError` on `os.open` (dir-fd
open, e.g. Windows) and on `os.fsync` (fs that reject dir-fsync) — both scoped narrowly, so a
`KeyboardInterrupt`/`MemoryError`/etc. still propagates; nothing broader is swallowed. Since
this step only runs after `os.replace` already succeeded, there's no continuation path where a
swallowed error here masks a write that didn't actually happen — worst case is exactly what
the docstring promises: best-effort hard durability, same risk profile as before this PR
existed. `test_atomic_write_fsyncs_parent_dir_after_replace` asserts the ordering
(`replace` index < `fsync_dir` index) and `test_atomic_write_dir_fsync_is_fail_open` proves a
forced `os.open` failure doesn't crash the write. Verified.

**2. `archive_assets` manifest-before-move — correct, traced the partial-move restore path.**
`entries` (name + type) is computed from the still-in-place assets, the manifest is written via
`atomic_write_text` (itself now durable per #1) *before* the move loop starts. Walked the
crash-mid-move case by hand against `restore_assets`: if `.claude` moved and the process dies
before `CLAUDE.md` moves, the manifest already names both. On restore: the `.claude` entry has
`src.exists()` (in the archive) → moves back normally. The `CLAUDE.md` entry has
`src.exists()` False (it never left the repo root) → `continue`s via the "already restored /
member absent" branch — which is exactly correct here since `CLAUDE.md` was never absent from
the root in the first place, so nothing needs to happen to it. No spurious conflict is raised
(the `dst.exists()` conflict check is only reached after the `src.exists()` check, so an
untouched original is never mistaken for "currently present, must overwrite"). End state is
byte-identical to pre-archive. `test_archive_writes_manifest_before_moving_and_is_recoverable`
confirms this with real fault injection (crash on the 2nd `shutil.move`) plus a full
byte-for-byte snapshot equality check after restore — good test, not just asserting the
manifest shape.

**3. `no_backup_litter` diff against `ctx.before` — correct baseline, correct direction.**
Traced `runner.run_cell`: `before = snapshot(target)` is taken unconditionally before every
install, for every cell, so `ctx.before` is always populated when this metric runs (not
metric-conditional). `snapshot()` keys are posix relpaths under the workdir root, which is
exactly what `p.relative_to(ctx.workdir).as_posix()` produces — same coordinate space, so the
membership check is valid. Direction is right: `.bak*` paths already in `before` (checked into
the fixture pre-install) are excluded; anything new is still counted. Matches the #101 false
positive (fixture's own `errors.jsonl.bak.*`) without hiding an installer-created backup.

**4. Scope discipline — deferrals are correctly redesign-sized, not cop-outs, with one
tracking gap.** Read the #149 thread: finding 1 (in-place settings merge isn't
manifest-reversible) requires changing the amend *contract* itself, and B10 baseline-mode
requires a whole new grading mode across the metric set — both genuinely bigger than a fix-in-
place, and both are named with a concrete next step in the triage comment. Agreed, correctly
deferred. One gap I'd like closed before this is fully "filed": the original botfarm triage
comment's **finding 6** ("amend path silently leaves stale config module lists" —
`config_module_lists_complete` failure) isn't mentioned as fixed, deferred, or explicitly
out-of-scope in the follow-up triage comment on #149, and I don't see a separate issue for it.
Not a defect in this PR (it's pre-existing amend behavior, orthogonal to atomicity/durability),
but as written it's quietly dropped rather than tracked — please add a line to the #149
comment (or spin a stub) so it doesn't get lost.

Must-fix: None

Tech-debt: Finding 6 from the original #149/botfarm triage comment ("amend path silently
leaves stale config module lists," `config_module_lists_complete` failure) isn't carried
forward into the follow-up triage comment as fixed/deferred/out-of-scope — add a tracking line
or stub issue so it isn't silently dropped from the #149 backlog.

parametrization:
Requestor: Tariq.Morales
Requestee: Ibrahim.El-Amin
RequestOrReplied: Replied

**Review: Must-fix resolved — the parent-dir fsync now has a load-bearing, non-tautological regression test; clears QA**

**AMENDED** (was ChangesRequested). Ibrahim pushed `ebb553c` adding `test_atomic_write_really_fsyncs_the_parent_directory`. Independently re-verified at PR HEAD `ebb553c` — did not trust the report:

- Read the new test: it records the parent directory's inode (`os.stat(parent).st_ino`), spies on `os.fsync` capturing `os.fstat(fd).st_ino` for every fsynced fd, and asserts `parent_ino in fsynced_inodes`. That correlates on the **dir fd specifically** (by inode), not merely "some fsync fired" — the temp-file fsync has a different inode and does not satisfy the assertion.
- Re-ran MY mutation: gutted the `os.fsync(dir_fd)` line inside `_fsync_dir` (kept `os.open`/`os.close` real). The new test now **FAILS** — `parent-dir fd was never fsynced (durability syscall dropped)`, asserting `872239 in [872240]` (only the temp file's inode was fsynced). Under this exact mutation the whole suite passed before; now it's caught. Restored the line → the test passes and the full suite is green (466 passed, +1 vs the prior 465). Genuine pass→fail→pass, non-tautological.
- Skip guard: `_dir_fsync_supported()` mirrors the code's own fail-open path (open+fsync a dir fd, `OSError` → skip), so runners without dir-fsync support skip cleanly rather than fail — CI stays green on Windows / unsupported filesystems.

Gap closed.

Must-fix: None

Tech-debt:
1. `_fsync_dir`'s `finally: os.close(dir_fd)` still isn't guarded by `try/except` — a (rare) close-time `OSError` would propagate past the "never crash the install" fail-open contract. Low priority (O_RDONLY dir-fd close essentially never fails); a one-line guard would make the fail-open promise complete.
2. #149's triage comment doesn't name finding 6 (stale config module lists on amend) explicitly among the deferred items — confirm it's folded into the B10 baseline-mode bucket, or give it its own line so it doesn't read as dropped.

parametrization:
@tariq-morales-qa Must-fix addressed — pushed in `ebb553c`.

**New test:** `test_atomic_write_really_fsyncs_the_parent_directory` (in `framework/tests/test_repo_space.py`). It spies on `os.fsync`, records the inode behind every fsynced fd via `os.fstat(fd).st_ino`, runs `atomic_write_text`, and asserts the **parent directory's** inode (`os.stat(parent).st_ino`) is among the fsynced inodes — i.e. the assertion is on the dir-fsync specifically, not "some fsync happened." It skips cleanly on platforms that can't fsync a directory fd (same fail-open guard the code uses), so CI stays green on all runners.

**Non-tautology proof (as you'd re-check it):** with the new test in place I temporarily gutted the `os.fsync(dir_fd)` line inside `_fsync_dir` (kept `os.open`/`os.close` real, exactly your mutation) → the test **FAILED**:
```
AssertionError: parent-dir fd was never fsynced (durability syscall dropped)
assert 228596 in [228597]
```
(228596 = parent dir inode, 228597 = temp-file inode — only the temp file was fsynced.) Restoring the line → passes. So a silent drop of the durability syscall now fails a test.

Full suite: **466 passed** (was 465, +1). `ruff check framework/` clean.

Ready for re-verify — over to you to amend the verdict.

PR #169, opened 2026-07-06
harness(tech-debt): enforce lifecycle kickoff-persistence — fail loud if state.json does not advance current_wave

parametrization:
## Summary

Wave 1 ran its **entire life un-stamped** — `state.json` stuck at `wave-5` — because the
kickoff transition's state write silently never landed, discovered only at wrapup and
requiring a full retroactive `allocate→start→scope→kickoff→wrapup` reconciliation. This
adds a guard so a wave can no longer proceed un-stamped:

- **`lifecycle.py`**: new pure `kickoff_persisted(state, wave) -> (ok, message)` — checks
  the three facts `kickoff_wave` persists together (`current_wave` pointer, `wave_{W}_active`,
  `wave_{W}_kicked_off_at`). Exposed as a new CLI sub-command `wave assert-kickoff <W>
  [--state PATH]` (exit 0 + message on success, exit 1 + `ERROR:` on stderr on failure).
- **wave-start skill** (both `.claude/skills/wave-start/SKILL.md` and
  `framework/assets/skills/wave-start/SKILL.md` — #116 dual-deploy parity): step 7's kickoff
  now runs `wave assert-kickoff "$W"` immediately after `wave kickoff` and `exit 1`s with a
  STOP message if it fails, instead of silently continuing.

### Where the guard lives and why
Put the check in `lifecycle.py` as a pure `state -> (bool, str)` function rather than only
in the skill's bash, so it's directly unit-testable (no subprocess/tmp-file plumbing needed
for the core logic) and reusable from any kickoff-driving surface. The CLI sub-command is a
thin wrapper the skill (or any future caller) shells out to.

### Load-bearing test
`test_assert_kickoff_persisted_traps_the_wave1_failure` (+ 3 companion tests) in
`framework/tests/test_lifecycle.py`. Verified by hand: patching the guard's `if problems:`
branch to `if False and problems:` (simulating a reverted/no-op guard) breaks
`test_assert_kickoff_persisted_traps_the_wave1_failure`,
`test_assert_kickoff_persisted_false_when_active_flag_missing`, and
`test_cli_assert_kickoff_exit_codes` (3 of 4 new tests fail); reverting the patch makes all
4 pass again. `test_wave_start_gates_kickoff_on_assert_kickoff` in
`test_wave_skill_lifecycle_wiring.py` additionally proves the skill wiring (guard present,
placed after `wave kickoff`, and actually `exit 1`s) in **both** dual-deploy copies.

### Safety
No mutation of the real `.claude/state.json` — all tests operate on `tmp_path` fixture state
files or in-memory dicts.

## Test plan
- [x] `ruff check framework/` — all checks passed
- [x] `ENVIRONMENT=test python -m pytest framework/tests/` — 472 passed (up from 466 baseline)
- [x] Manual revert check: broke 3/4 new tests when the guard was neutered, all pass restored

Closes #168

Co-Authored-By: Nia Rossi <Nia.Rossi@gmail.com>

### DIFF
--- a/.claude/skills/wave-start/SKILL.md
+++ b/.claude/skills/wave-start/SKILL.md
@@ -159,13 +159,23 @@ go-ahead. **Only after** the User approves, record the kickoff transition live:
 
 ```bash
 python3 "$LIB/lifecycle.py" wave kickoff "$W" --merge-model "$MERGE_MODEL"
+
+# Guard (#168): fail loudly if the write above did not actually land — the exact Wave 1
+# failure was a wave running its entire life un-stamped (state.json stuck at the prior
+# wave) because this step silently didn't persist. Do NOT proceed past a non-zero exit.
+python3 "$LIB/lifecycle.py" wave assert-kickoff "$W" || {
+  echo "STOP: kickoff did NOT persist to state.json — the wave cannot proceed un-stamped."
+  echo "Re-run the kickoff command above, then re-check with 'wave assert-kickoff' before continuing."
+  exit 1
+}
 ```
 
 This stamps `wave_{W}_kicked_off_at`, declares `wave_{W}_merge_model`, and re-points
 `current_wave`. Confirm the model with the User first: a wave whose per-issue PRs base on
 the integration branch is `wave-branch`; one whose PRs base straight on the default branch
 is `direct-to-main`. Do **not** run this before the go-ahead — it marks the wave as
-officially started and work beginning.
+officially started and work beginning. The `assert-kickoff` guard immediately after is not
+optional — it is what makes the un-stamped-wave failure mode impossible.
 
 ### 8. Report
 
@@ -187,3 +197,5 @@ officially started and work beginning.
   re-runs
 - Wave kickoff (Step 7) is an approval gate: the plan is presented and the `wave kickoff`
   lifecycle transition fires only on the User's explicit go-ahead
+- Kickoff persistence is enforced, not manual: `wave assert-kickoff` (#168) fails the step
+  loudly if `current_wave` did not actually advance — no more silent un-stamped waves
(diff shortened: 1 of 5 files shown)


parametrization:
Requestor: Tariq.Morales
Requestee: Nia.Rossi
RequestOrReplied: Replied

**Review: kickoff-persistence guard — checks exactly the three facts kickoff_wave writes, mutation-checked load-bearing**
Must-fix: None
Tech-debt: None

Verification performed:
- CI: `gh pr checks 169` — all 11 checks green, re-confirmed.
- Checked out `N.Rossi/0168-lifecycle-kickoff-persistence` in my worktree.
- Correctness: compared `kickoff_persisted`'s three checked facts (`current_wave` pointer, `wave_{W}_active`, `wave_{W}_kicked_off_at`) against `kickoff_wave`'s actual write set — they match exactly (the optional `merge_model` key is correctly excluded from the required check, since it's conditional). No stale-flag risk across waves either: `wave_{W}_active` is wave-id-scoped (not a generic "any wave active" flag) and is flipped to `False` at `wrapup_wave`, so a prior wave's leftover flag can't false-pass a new wave's guard.
- Mutation check on the load-bearing claim: patched `if problems:` to `if False and problems:` in `kickoff_persisted` (neutering the guard to always report success, simulating a reverted/no-op guard). Ran the 4 new lifecycle tests + the skill-wiring test: 3 of 4 new tests failed exactly as claimed (`test_assert_kickoff_persisted_traps_the_wave1_failure`, `test_assert_kickoff_persisted_false_when_active_flag_missing`, `test_cli_assert_kickoff_exit_codes`); the 4th (`test_assert_kickoff_persisted_true_after_real_kickoff_wave`) and both `test_wave_start_gates_kickoff_on_assert_kickoff` skill-wiring parametrizations still passed — expected, since those only exercise the success path / static bash wiring, not the guard's failure-detection logic. Restored `if problems:`: all 29 tests in `test_lifecycle.py` + `test_wave_skill_lifecycle_wiring.py` green again, clean diff.
- Skill wiring: `diff .claude/skills/wave-start/SKILL.md framework/assets/skills/wave-start/SKILL.md` → byte-identical; the guard is placed immediately after `wave kickoff` in both copies and `exit 1`s with a STOP message on failure, matching the PR's description.
- Safety: confirmed all new/changed tests operate on `tmp_path` fixtures or in-memory dicts — no mutation of the real `.claude/state.json`.
- Full suite: `ENVIRONMENT=test python -m pytest framework/tests/` → 472 passed (matches PR description). `ruff check framework/` → clean.

PR #171, opened 2026-07-06
fix(trust-signals): durable review-catch ledger survives verdict amend-in-place (#164)

parametrization:
## Summary

`trust_signals.extract_signals` used to re-derive `must_fix_caught`/`must_fix_received`
purely by parsing each PR's **current** comment bodies. The charter's amendment convention
has a reviewer edit their blocking `Must-fix` comment **in place** to `Replied`/`Must-fix:
None` once the fix lands, so the historic catch silently vanished — and a reviewer who
authored no PRs of their own dropped out of the engineer map **entirely**. This is exactly
what happened to Tariq in Phase 6 Wave 6 (`wave_6_counter_corrections` in `state.json`):
`must_fix_caught=0`, hand-reconstructed at retro from PR timelines.

## Design decision: option (b) — durable issue-time ledger

Chosen over the two other directions named in #164/#165:
- (a) reading GitHub comment edit-history — rejected: adds a new SCM/API dependency to what
  is currently a pure scorer.
- (c) a distinct `Replied` follow-up instead of amending in place — rejected: changes the
  charter convention, two comments per cycle, and doesn't fix history that's already amended.

**Chosen (b):** `record_review_catch()` durably appends one entry per changes-requested
verdict to `wave_{wave}_review_catches` in the project state file **at issue time** — the
moment the blocking verdict comment is posted, before any later amendment can erase it — via
`lifecycle.persist` (preserves the file's compact-inline shape; validates JSON before/after).

`extract_signals` now reads this ledger (`_review_catches_for_wave`) and, for any PR with
≥1 recorded catch, treats it as **authoritative** for that PR's changes-requested accounting
instead of re-deriving it from the (possibly already-amended) live comment body. A PR with no
ledger entry falls back to the legacy live-comment-parsing path unchanged — every existing
wave/project keeps scoring exactly as before (zero regression risk for waves that never call
the new recorder).

The per-PR accounting loop is extracted out of `extract_signals` into a shared `_account_pr`
helper so `extract_signals` and the tests exercise the identical logic (no drift between what
ships and what's tested). A `record-catch` CLI subcommand is the concrete issue-time call
site (`trust_signals.py record-catch <wave> --repo O/N --pr P --requestor R --requestee E`).

## Files touched
- `framework/assets/lib/trust_signals.py` — `_review_catches_for_wave`, `record_review_catch`,
  `_account_pr` (extracted + ledger-aware), `extract_signals` (wired to the ledger),
  `record-catch` CLI subcommand, updated module docstring.
- `framework/tests/test_trust_signals.py` — 14 new tests (ledger read/write round-trip,
  ledger-authoritative accounting, no-double-count guard, repo/pr-scoping guard, CLI,
  end-to-end `extract_signals` regression, and a synthetic Wave-6 fixture reproducing the
  real #154/#156/#160/Tariq scenario from `wave_6_counter_corrections`).

## Load-bearing test
`test_extract_signals_amend_in_place_does_not_erase_catch` — exercises the real
`extract_signals` entrypoint end to end (the same one `extract`/`score` CLI subcommands call)
against an amended-in-place comment body, with the catch pre-recorded in the ledger.
Verified load-bearing by reverting the fix (forcing `_account_pr`'s ledger-derived `catches`
list to always be `[]`) and re-running: this test **and** its Wave-6-fixture sibling
(`test_wave6_fixture_rescoring_with_amended_catches`) both fail with the exact #164 symptom —
`must_fix_received` drops to 0 and `"Tariq Morales"` `KeyError`s out of the signals map
entirely. Reverted the revert before committing.

## Acceptance criteria
- [x] A wave with N amended must-fix cycles scores `must_fix_caught=N` for the reviewer and
      `must_fix_received` for the authors — proven by the 3-PR Wave-6 fixture.
- [x] The Wave-6 data re-scores correctly as a fixture (`test_wave6_fixture_rescoring_with_amended_catches`).
- [x] Aggregate `changes_requested_cycles` stays consistent — `sum(s.rework_cycles for s in
      sigs.values()) == 3`, matching the wrapup-recorded `wave_6_changes_requested_cycles: 3`.
- [x] Load-bearing (revert→fail) test proving the amend-then-rescore path.

## Dual-deploy (#116)
Not applicable — confirmed `ls .claude/lib` has no runtime copy in this repo; `framework/assets/lib`
is the single canonical location for this module.

## CI
- `ruff check framework/` → all checks passed
- `ENVIRONMENT=test python -m pytest framework/tests/` → 480 passed (was 466; +14 new tests)

Closes #164

### DIFF
--- a/framework/assets/lib/trust_signals.py
+++ b/framework/assets/lib/trust_signals.py
@@ -26,6 +26,35 @@ Two cleanly separated layers:
     is a pure function so the model is unit-testable, and usable standalone on a
     hand-built signal dict, without touching the SCM at all.
 
+Durable review-catch ledger (#164)
+-----------------------------------
+The charter's verdict-amendment convention has a reviewer edit their blocking
+``Request``/``Must-fix:`` comment **in place** to ``Replied``/``Must-fix: None``
+once the fix lands. Because extraction used to derive ``must_fix_caught`` /
+``must_fix_received`` purely by re-parsing the PR's *current* comment bodies,
+an amendment silently erased the historic catch — the reviewer's signal
+vanished and, if they authored no PRs of their own, they dropped out of the
+engineer map entirely (Phase 6 Wave 1: the wave's standout reviewer scored
+``must_fix_caught=0``).
+
+Fix (design option (b) from #164/#165 — chosen over reading GitHub comment
+edit-history or a distinct ``Replied``-follow-up convention change: it needs no
+new SCM/API dependency and keeps the signal durable at the moment it is
+created): :func:`record_review_catch` durably appends one entry per
+changes-requested verdict to ``wave_{wave}_review_catches`` in the project
+state file, **at issue time** — i.e. the moment the blocking verdict comment is
+posted, before any later amendment can erase it. :func:`extract_signals` reads
+this ledger (:func:`_review_catches_for_wave`) and, for any PR that has >=1
+recorded catch, treats the ledger as **authoritative** for that PR's
+changes-requested accounting instead of re-deriving it from the (possibly
+already-amended) live comment body. A PR with no ledger entry falls back to the
+legacy live-comment-parsing path unchanged, so projects/waves that predate this
+ledger keep scoring exactly as before.
+
+CLI: ``trust_signals.py record-catch <wave> --repo O/N --pr P --requestor R
+--requestee E`` — meant to be invoked immediately after posting a blocking
+verdict comment (see the module CLI section below).
+
 Signals per engineer (all integers, all countable from the merged-PR set):
 
   ===========================  ============================================
@@ -54,6 +83,12 @@ CLI:
                                                                 # proposed deltas
                                                                 # + forced
                                                                 # negative line
+  trust_signals.py record-catch <wave> --repo O/N --pr P --requestor R
+                                 --requestee E [--at TS] [--status PATH]
+                                                                # durably record
+                                                                # one changes-
+                                                                # requested catch
+                                                                # at issue time
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


parametrization:
Requestor: Tariq.Morales
Requestee: Paloma.Gupta
RequestOrReplied: Replied

**Review: durable review-catch ledger — mutation-checked load-bearing, legacy path verified byte-for-byte equivalent**
Must-fix: None
Tech-debt: `_account_pr`'s ledger branch increments `must_fix_received`/`must_fix_caught` once per ledger entry but only bumps `author_sig.rework_cycles` once per PR (flag-gated), same as the legacy path — consistent, not a bug, but worth a one-line docstring note on `record_review_catch` since its "N distinct changes-requested cycles" phrasing could read as implying N `rework_cycles`, which it doesn't.

Verification performed:
- CI: `gh pr checks 171` — all 11 checks green, re-confirmed.
- Checked out `P.Gupta/0164-trust-signals-amend-scoring` in my worktree.
- Mutation check on the load-bearing claim: forced `_account_pr`'s `catches` list to always be `[]` (no-op'ing the ledger branch, equivalent to reverting the fix). Both `test_extract_signals_amend_in_place_does_not_erase_catch` and `test_wave6_fixture_rescoring_with_amended_catches` failed with the exact claimed symptom — `KeyError: 'Tariq Morales'` (the reviewer drops out of the signals map entirely because their only credit came via `must_fix_caught` on a PR whose live comment is now clean). Restored the real filter and re-ran: both pass, full suite back to 480/480 green, `ruff check framework/` clean, no diff left in the tree. Non-tautological, confirmed.
- Read the "zero regression risk" claim for real: compared the new `_account_pr`'s `else` branch (empty ledger) line-by-line against the old inline loop it replaced in `extract_signals` — identical logic (same verdict iteration, same `must_fix_received`/`must_fix_caught`/`rework_cycles`/`review_false_positives` accounting), just split so the false-positive loop runs unconditionally. Confirms a wave/project that never calls `record_review_catch` truly scores unchanged.
- Confirmed no double-count risk: when a PR has >=1 ledger entries, the live-comment `verdicts` are still parsed but only consulted for `false_positive` retractions, never for `changes_requested` — so ledger and legacy paths can't both fire for the same PR's must-fix accounting.
- Confirmed dual-deploy claim: `ls .claude/lib` → no such directory in this repo, so `framework/assets/lib/trust_signals.py` is genuinely the single canonical copy (no runtime mirror to miss).
- Full suite: `ENVIRONMENT=test python -m pytest framework/tests/` → 480 passed (matches PR description). `ruff check framework/` → clean.

PR #172, opened 2026-07-06
process: hard pre-review gate — new behavior needs a load-bearing test (#167)

parametrization:
## Summary
- New hard, fail-CLOSED PreToolUse hook `require_load_bearing_test` (S2, #167): blocks
  `gh pr create`/`gh pr ready` when the diff adds a substantive line to a "behavior" `.py`
  file with no accompanying test-file change in the same diff, or when the diff itself
  can't be verified (network/API failure, unresolvable `--repo`/`--head`).
- Bypass only via a validated `LOAD_BEARING_TEST_EXCEPTION=<class>:<rationale>` naming a
  class configured under `policy.load_bearing_test_exceptions` (empty by default — no
  bypass exists in this repo until a class is configured), mirroring the existing
  `ADMIN_MERGE_EXCEPTION` convention in `validate_pr_ci_status.py`.
- Wired into all four `hooks.pre_bash` sync points (`framework.config.schema.json`,
  `_framework_config._DEFAULTS`, `bootstrap._schema_defaults()`,
  `framework.config.example.json`) plus this repo's live `.claude/framework.config.json`.
- New charter section `charter/pull-requests.md § Pre-Review Self-Check`, plus a
  `charter/hooks.md` enforcement-map row — canonical (`framework/assets/team/charter/**`)
  and live (`.claude/team/charter/**`) copies both updated (#116).
- Load-bearing test: `framework/tests/test_require_load_bearing_test.py` (18 tests).
  Verified by manually neutering the gate's core conditional
  (`if not behavior_files or test_touched: return None` → `if True: return None`) and
  re-running the suite: 6 tests failed, including the flagship
  `test_blocks_new_behavior_without_test_file` — confirming the tests are load-bearing,
  not tautological. Reverted the neutering before committing.

## Related Issues
Closes #167

## Review Checklist
- [ ] Reviewed by another team member
- [ ] Must-fix items resolved
- [ ] Tech debt items filed as GitHub Issues (if any)

PR #182, opened 2026-07-06
fix(dispatcher): fail-closed top-level guard around hook check() (#175)

parametrization:
## What

Closes #175. Two related crash-safety gaps in the shared hook plumbing, both flagged in the #172 review:

1. **`dispatcher.py` swallowed any uncaught `check()` exception as ALLOW.** For most hooks (documented fail-open matchers) that's the intended posture — a parse bug shouldn't block the user's unrelated command. But `require_load_bearing_test` (#167) is deliberately **fail-closed**: its whole job is to stop unverified new behavior from reaching review, so a crash inside it must BLOCK, not silently pass the exact thing it exists to catch. Before this PR, a bug in that hook (or any future fail-closed hook) would have been indistinguishable from "everything's fine."
2. **`_added_substantive_lines(patch)`** in `require_load_bearing_test.py` assumed `patch` was a string once truthy, with no `isinstance` guard — an unexpected JSON shape from the compare API (e.g. a non-string `patch` field) would raise `AttributeError` on `.splitlines()`, which (per gap 1) used to be swallowed as allow.

## Design: per-hook fail-direction declaration

There was no existing mechanism for a hook to declare its fail-direction, so this PR adds a minimal one: a module-level **`FAIL_OPEN = True`** attribute.

- **Declared fail-open** (`FAIL_OPEN = True`): a crash in `check()` is skipped exactly as before — the dispatcher moves on to the next hook. Added to the 9 hooks in the default `hooks.pre_bash` list whose docstrings already document a fail-open posture (`block_no_verify`, `block_git_config`, `no_worktree_self_delete`, `warn_zsh_wordsplit`, `validate_labels`, `validate_review_comment_format`, `validate_workflow_paths_coverage`, `validate_pr_ci_status`, `block_squash_wave_merge`) — their existing behavior is unchanged, just now explicit and auditable.
- **Undeclared** (no `FAIL_OPEN` attribute — the default): a crash now **blocks**, with a reason naming the hook and pointing at the fix (either the underlying bug, or add `FAIL_OPEN = True` if the hook really is meant to fail open). `require_load_bearing_test` deliberately does **not** set `FAIL_OPEN` — it's the repo's one fail-closed hook by design (#167), so "default to block" is exactly the behavior it needs, and any *future* hook that forgets to declare a fail-direction is safe-by-default rather than silently fail-open.

This flips the dispatcher's implicit default from "swallow every crash" to "block unless a hook opts into fail-open" — the least invasive change that closes the actual gap (an undeclared/forgotten-to-declare hook is now the safe case, not the dangerous one) while leaving every currently-fail-open hook's behavior bit-for-bit identical (via explicit opt-in).

The crash-block path also logs via `_framework_log.log_pretooluse_block` (audit trail, same as other block decisions) and is suppressed under `ENVIRONMENT=test`/pytest like all other event-log writes.

Scope note: this only touches `dispatcher.py` (the PreToolUse Bash/Agent gate), matching the issue. `start_dispatcher.py` / `post_dispatcher.py` / `stop_dispatcher.py` have their own "always allow, never block" contracts (Stop in particular must never block by design) and are out of scope.

## Testing

- `framework/tests/test_dispatcher_fail_closed.py` (new, 5 tests) drives `dispatcher.main()` in-process:
  - **Load-bearing** `test_undeclared_hook_crash_blocks`: a hook without `FAIL_OPEN` that raises → dispatcher exits 2 with a fail-closed reason naming the hook. Verified this fails (asserts `0 == 2`) when the `except Exception: continue` guard is reverted.
  - `test_declared_fail_open_hook_crash_still_allows`: a `FAIL_OPEN = True` hook that raises → exit 0, nothing printed (pre-#175 behavior preserved).
  - `test_mixed_order_fail_closed_hook_still_blocks_after_fail_open_hook`: a crashing fail-open hook earlier in the list doesn't short-circuit the loop for a later hook.
  - **Load-bearing** `test_added_substantive_lines_rejects_non_str_patch`: `None`/`int`/`list`/`dict` patch values all return `[]` instead of raising. Verified this fails with `AttributeError: 'int' object has no attribute 'splitlines'` when the `isinstance` guard is reverted.
  - `test_added_substantive_lines_still_parses_real_patch`: sanity check the real parsing path is untouched.
- Updated `framework/tests/test_agent_stop_dispatch.py`: the old `test_agent_fail_open_on_import_error_and_raise` asserted the *previous* (now-intentionally-changed) behavior — a raising Agent-gate hook used to be swallowed. Split into `test_agent_missing_module_skipped_undeclared_raise_blocks` (new default: blocks) and `test_agent_declared_fail_open_raise_still_allows` (opt-out preserved), both exercised through the real bootstrap-installed dispatcher via subprocess.
- Also refreshed `charter/hooks.md` (canonical + this repo's rendered copy) to describe the new per-hook `FAIL_OPEN` opt-in instead of "fail open by default" without qualification.

**Test count:** 507 (v0.7.0 baseline) → 513 (+6: 5 new in `test_dispatcher_fail_closed.py`, +1 net in `test_agent_stop_dispatch.py` after the 1-test split into 2).

## Verification

- `ENVIRONMENT=test python -m pytest framework/tests/` → 513 passed
- `ruff check framework/` → all checks passed
- `python3 framework/install/reinstall.py --check` → in sync (hooks/lib run canonical-by-reference in this repo — `.claude/settings.json` points straight at `framework/assets/hooks/*`, so there's no copied `.claude/hooks/` twin to drift for this change)
- `python3 framework/install/manifest.py --check` → in sync (no new installed files)

This satisfies W2-retro proposal #1 (prerequisite to trusting the #167 gate) — `require_load_bearing_test` can now actually be trusted to fail closed end-to-end, including on its own bugs.

### DIFF
--- a/.claude/team/charter/hooks.md
+++ b/.claude/team/charter/hooks.md
@@ -9,7 +9,12 @@ reviewer count, identity roster, …) from that same config. All hooks are
 stdlib-only, and fail open by default — a broken hook never blocks unrelated
 work — with one deliberate exception: `require_load_bearing_test` (see below and
 [pull-requests.md § Pre-Review Self-Check](pull-requests.md)) fails **closed** on
-the review-request path, by owner-approved design (#167).
+the review-request path, by owner-approved design (#167). "Fail open" is now an
+explicit, per-hook opt-in (`FAIL_OPEN = True`) rather than the dispatcher's
+implicit default: if a hook's `check()` raises an uncaught exception, the
+dispatcher blocks unless that module declared itself fail-open, so a hook that
+forgets to declare (like `require_load_bearing_test`, deliberately) fails closed
+on its own crash too (#175).
 
 ## Rule → Enforcement Map
 
--- a/framework/assets/hooks/block_git_config.py
+++ b/framework/assets/hooks/block_git_config.py
@@ -50,6 +50,11 @@ from _shell_parse import (  # noqa: E402
     tokenize,
 )
 
+#: Declares this hook's fail-direction to the dispatcher (#175): an uncaught
+#: exception from `check()` allows the command, matching this module's
+#: documented "parse failure -> fail open" posture.
+FAIL_OPEN = True
+
 # Config keys the policy actually protects (commit identity). The whole `user.`
 # namespace is treated as identity (user.name / user.email / user.signingkey).
 # Non-identity / operational keys (core.hooksPath, commit.gpgsign, …) are NOT
--- a/framework/assets/hooks/block_no_verify.py
+++ b/framework/assets/hooks/block_no_verify.py
@@ -48,6 +48,11 @@ from _shell_parse import (  # noqa: E402
     tokenize,
 )
 
+#: Declares this hook's fail-direction to the dispatcher (#175): an uncaught
+#: exception from `check()` allows the command, matching this module's
+#: documented "parse failure -> fail open" posture.
+FAIL_OPEN = True
+
 
 def _segment_has_no_verify(segment: list[str]) -> bool:
     """True if a tokenized git commit/push segment carries --no-verify or -n.
--- a/framework/assets/hooks/block_squash_wave_merge.py
+++ b/framework/assets/hooks/block_squash_wave_merge.py
@@ -49,6 +49,11 @@ from _shell_parse import (  # noqa: E402
     walk_flag_values,
 )
 
+#: Declares this hook's fail-direction to the dispatcher (#175): an uncaught
+#: exception from `check()` allows the command, matching this module's
+#: documented "base unresolvable -> fail-open" posture.
+FAIL_OPEN = True
+
 
 def _integration_regex(template: str) -> re.Pattern[str]:
     """Build an anchored regex from a ``branch.integration`` template.
--- a/framework/assets/hooks/dispatcher.py
+++ b/framework/assets/hooks/dispatcher.py
@@ -22,8 +22,27 @@ Exit codes:
   0 — allow (all passed, or aggregated warnings)
   2 — block (first blocking hook wins)
 
-Fail-safe posture: a missing module is skipped; a hook that raises is skipped
-(never let one hook crash the whole gate).
+Fail-safe posture, per hook (#175)
+====================================
+
+A missing/uninstalled module is always skipped (`ImportError` — nothing to run).
+
+A hook whose `check()` raises an uncaught exception is a different story: most
+hooks in this repo are documented fail-open (a matcher that can't parse a
+command should not block the user over its own bug), but at least one —
+`require_load_bearing_test` (#167) — is *deliberately* fail-CLOSED: its entire
+purpose is to stop unverified new behavior from reaching review, so an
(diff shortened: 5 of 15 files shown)


parametrization:
Requestor: Tariq.Morales
Requestee: Nia.Rossi
RequestOrReplied: Replied

**Review: fail-closed dispatcher guard is correctly scoped, load-bearing, and doesn't wrongly flip any hook's fail-direction**
Must-fix: None
Tech-debt:
- `validate_commit_identity` is in `hooks.pre_bash` (11 modules total, `identity.enforce: true` in this repo's config) and is one of only 2 modules left undeclared (the other is `require_load_bearing_test`). Its own docstring documents a deliberate fail-closed security posture ("On a tokenize failure it does NOT fail open... it is BLOCKED"), so leaving it undeclared is the *correct* call and matches the new default — but the PR body/dispatcher docstring only calls out `require_load_bearing_test` as the intentional fail-closed exception. A one-line mention of `validate_commit_identity` alongside it would save the next reviewer from having to re-derive that this is by design rather than an oversight.
- Saw one non-reproducible failure in `test_meta_install_aggregate_is_idempotent` (unrelated ontology-freshness test, not touched by this diff) on a single full-suite run; two subsequent full runs were clean (513 passed). Looks like pre-existing order/environment flakiness, not caused by this change — flagging for someone to keep an eye on, not blocking here.

Verification performed:
- Enumerated all 11 `hooks.pre_bash` modules against the 9 `FAIL_OPEN = True` declarations added: `block_no_verify`, `block_git_config`, `no_worktree_self_delete`, `warn_zsh_wordsplit`, `validate_labels`, `validate_review_comment_format`, `validate_workflow_paths_coverage`, `validate_pr_ci_status`, `block_squash_wave_merge`. The 2 undeclared modules (`require_load_bearing_test`, `validate_commit_identity`) are both legitimately fail-closed by their own documented design — no hook that needs to stay fail-open is missing the declaration.
- Confirmed `post_dispatcher.py` / `start_dispatcher.py` / `stop_dispatcher.py` are untouched and remain unconditionally fail-open/advisory by contract (out of scope per PR description) — correct, since `hooks.post_bash`/`post_file`/`session_start`/`stop` hooks were never in play for this fix.
- `ENVIRONMENT=test python -m pytest framework/tests/` → 513 passed (clean on repeat runs); `ruff check framework/` → clean; `reinstall.py --check` and `manifest.py --check` → both exit 0.
- Mutation-checked both claimed load-bearing tests by reverting the production line and confirming the specific test fails, then restored:
  - Reverted the dispatcher's `except Exception as exc: ... FAIL_OPEN ... else block` back to bare `except Exception: continue` → `test_undeclared_hook_crash_blocks` fails (`assert 0 == 2`), all else in the file still passes. Confirms the test isn't tautological.
  - Reverted `if not patch or not isinstance(patch, str)` back to `if not patch` in `_added_substantive_lines` → `test_added_substantive_lines_rejects_non_str_patch` fails with the exact `AttributeError: 'int' object has no attribute 'splitlines'` the PR describes.

PR #183, opened 2026-07-06
feat(gate): per-behavior-file load-bearing-test pairing + refactor exception class (#174, #176)

parametrization:
## Summary

Combined PR for Wave 3 S2 (#174) and S3 (#176) — both touch
`require_load_bearing_test.py`, the Wave 2 (#167) hard pre-review gate.

- **S2 (#174):** pairing is now checked **per behavior file**, not once for
  the whole diff. Previously, ANY substantive test-file change anywhere in a
  PR's diff satisfied the entire diff — a PR touching `hooks/foo.py` (new
  behavior) and `lib/bar.py` (new behavior) with a test change only for `foo`
  let `bar.py`'s new behavior through completely ungated. Each new-behavior
  file must now have its own paired test-file change.
- **S3 (#176):** `policy.load_bearing_test_exceptions` shipped empty, so a
  genuine pure-refactor PR (no new behavior — nothing to pair a test to) was
  hard-blocked with zero configured bypass. It now ships pre-seeded with one
  `refactor` class at rollout, mirroring the existing `ADMIN_MERGE_EXCEPTION`
  precedent's shape (map of `<class>` -> human rationale, validated via
  `<class>:<rationale>`).

## Granularity choice (S2): per-file, not per-directory

Chose **per-file** pairing. This repo's own behavior roots
(`framework/assets/hooks/`, `framework/assets/lib/`, `framework/install/`,
`framework/harness/`) all share ONE test root (`framework/tests/`) — a
per-directory bucket would still let module A's test satisfy module B's new
behavior whenever A and B share a behavior root, reproducing the exact
loophole #174 exists to close. Per-file is the only granularity that actually
closes it here.

A behavior file `X` counts as paired if the diff also substantively touches a
test file that is:

1. named `test_<stem>.py` / `<stem>_test.py` for `X`'s stem (the "mapped test
   path" — this repo's own convention, e.g. `hooks/foo.py` <-> `tests/test_foo.py`),
2. in the SAME DIRECTORY as `X` (for repos that co-locate tests beside
   source instead of a separate tree — this framework installs into many
   layouts), or
3. `conftest.py` (shared fixture infra isn't "for" one behavior file).

**Known trade-off:** rule 1 requires the `test_<stem>.py` naming convention.
Several of this repo's OWN existing hook/lib modules predate that convention
(e.g. `dispatcher.py`, `bootstrap.py` are covered today by tests with
differently-named files) and aren't covered by rule 2 or 3 either. The next
time one of those files gains genuinely new behavior, its PR will need a
matching test added/renamed, or use the documented `refactor` override if the
change is test-exempt. This is a deliberate policy tightening, not an
oversight — flagged here for visibility.

## Exception-class opt-in (S3)

`policy.load_bearing_test_exceptions` in `_framework_config._DEFAULTS`
(mirrored exactly in `framework.config.schema.json`'s `default` and in
`framework.config.example.json`) now ships:

```json
"load_bearing_test_exceptions": {
  "refactor": "Pure refactor / no external behavior change (renames, extraction, dead-code removal, formatting) - nothing new to test. State exactly what was restructured and confirm no observable behavior changed."
}
```

A PR opts in by prefixing its `gh pr create`/`gh pr ready` call:

```bash
LOAD_BEARING_TEST_EXCEPTION="refactor:<what you restructured, why behavior is unchanged>" gh pr create ...
```

`<rationale>` must be non-empty and specific; an unrecognized class or empty
rationale still blocks (fail-safe). Every attempt — authorized or not — is
logged to the events log for audit, same as before.

**Caveat documented in code:** dict-valued config keys merge ADDITIVELY over
`_framework_config._DEFAULTS` (`_deep_merge`), not by replacement. A repo
that writes `"load_bearing_test_exceptions": {}` in its own config does NOT
clear the seeded `refactor` class — it stays present. A repo wanting truly
zero classes would need to fork the runtime default. Out of scope for this
PR to fix (pre-existing `_deep_merge` semantics, used by other dict-valued
policy keys too); called out for visibility.

## Verification (bars from the story)

- **S2:** added `test_blocks_new_behavior_when_paired_test_is_for_an_unrelated_file`
  (new behavior in file X, test change only in unrelated file Y -> BLOCKED)
  and `test_blocks_only_the_specific_unpaired_behavior_file` (two new-behavior
  files, only one paired -> blocks on the unpaired one specifically). Manually
  confirmed load-bearing: monkeypatching `_behavior_file_paired` back to the
  old v1 "any test file touched satisfies everything" boolean makes the gate
  wrongly ALLOW the same diff (transcript kept during development).
- **S3:** added `test_seeded_refactor_class_allows_pure_refactor_with_no_repo_config`
  (fresh `tmp_path`, zero repo config, `refactor:` override -> ALLOWED against
  the real shipped `_DEFAULTS`) and `test_seeded_refactor_class_would_be_rejected_without_the_seed`
  (same override, `load_bearing_test_exceptions` resolved to `{}` -> BLOCKED).
  Manually re-verified against the real `_framework_config._DEFAULTS` object
  (zeroed it, confirmed block; restored it, confirmed allow).

## Checks

- `ENVIRONMENT=test python -m pytest framework/tests/` — 517 passed.
- `ruff check framework/` — all checks passed.
- `python3 framework/install/reinstall.py --check` — exit 0, already in sync.
  (No action needed: hooks are canonical-by-reference in this repo — no
  `.claude/hooks/` mirror exists yet, per the dogfood-pending state — and
  `team/charter/**` is an intentionally hand-evolved template, not
  reinstall-managed. Only `skills/` is byte-mirrored by `reinstall.py`, and
  this PR doesn't touch it, so no golden-manifest change is needed either.)

Closes #174
Closes #176


parametrization:
Requestor: Nia.Rossi
Requestee: Tariq.Morales
RequestOrReplied: Replied

**Review: per-file load-bearing-test pairing (#174) + seeded `refactor` exception class (#176) both do what they claim; green gate clean, both load-bearing claims mutation-verified**
Must-fix: None
Tech-debt:
- (a) Additive-merge caveat (dict-valued config keys merge over `_DEFAULTS` via `_deep_merge` rather than replacing): a repo writing `"load_bearing_test_exceptions": {}` cannot clear the seeded `refactor` class — confirmed empirically (`_deep_merge({"refactor": ...}, {})` still yields `{"refactor": ...}`). This doesn't defeat the gate's core guarantee — the bypass still requires an explicit, validated, audited `LOAD_BEARING_TEST_EXCEPTION=refactor:<rationale>` invocation, same enforcement posture as any other configured class — it only removes a repo owner's ability to fully opt out of this one default class. Pre-existing `_deep_merge` semantics shared by other policy dict keys, correctly called out as out-of-scope. Follow-up: give `_deep_merge` (or just this key) real delete/override semantics (e.g. a sentinel or `_replace` key) if a future repo actually needs to zero it out.
- (b) Stem-naming doesn't retroactively cover legacy modules (`dispatcher.py`, `bootstrap.py` etc. predate the `test_<stem>.py` convention and live in a different directory than their real tests, e.g. `test_agent_stop_dispatch.py`, `test_bootstrap_smoke.py`). Failure direction is over-blocking (false-positive BLOCK on a real, already-paired legacy module), not under-blocking — consistent with this hook's own documented fail-closed posture (#167) and self-resolves the next time one of those files touches (rename/add a conventionally-named test, or use the `refactor` override if genuinely exempt). Not a hole in the gate's guarantee.
- Minor observation (pre-existing, not introduced by this PR, not blocking): rule 3's blanket "`conftest.py` touch satisfies pairing for every behavior file in the diff" was already present pre-#174 (`test_conftest_and_test_dir_recognized_as_test` predates this PR) and is a narrower, deliberately-scoped survival of the old whole-diff-boolean loophole (any *unrelated* `conftest.py` edit still blankets-pairs every behavior file). Documented and intentional ("shared fixture infra isn't 'for' one behavior file"); flagging for awareness only, since it's the one remaining path by which #174's loophole could theoretically reopen.

**Verification performed**
- Green gate: `ENVIRONMENT=test python -m pytest framework/tests/` → 517 passed; `ruff check framework/` → clean; `reinstall.py --check` → in sync; `manifest.py --check` → in sync.
- S2 mutation check: reverted `check()`'s per-file pairing to the old v1 whole-diff boolean (`unpaired = [] if test_files else behavior_files`) — both new regression tests (`test_blocks_new_behavior_when_paired_test_is_for_an_unrelated_file`, `test_blocks_only_the_specific_unpaired_behavior_file`) then failed because the gate wrongly ALLOWED (returned `None` instead of `block`). Restored via `git checkout --`; suite re-passes (28/28 in that file).
- S3 mutation check: zeroed the seed (`"load_bearing_test_exceptions": {}` in `_framework_config._DEFAULTS`) — `test_seeded_refactor_class_allows_pure_refactor_with_no_repo_config` then failed (`block` instead of `allow`), confirming the seed itself (not the override mechanism generally) is load-bearing for that test. Restored; suite re-passes.
- Confirmed the per-FILE (not per-directory) granularity choice is correct for this repo: since all of this repo's behavior roots (`framework/assets/hooks/`, `framework/assets/lib/`, `framework/install/`, `framework/harness/`) share one separate test root (`framework/tests/`), the same-directory rule (2) essentially never fires here — closure depends on rule 1 (exact stem match) plus the narrower conftest.py exception (3), which is a materially smaller bypass surface than v1's "any test file anywhere" boolean. Per-directory bucketing would have reproduced the #174 loophole exactly as the PR argues; per-file genuinely closes it for this project's layout.

PR #185, opened 2026-07-06
feat(promotion): port promotion/genericization pipeline P0 from noorinalabs audit (#102)

parametrization:
## Summary

Ports the P0 flagship slice of the promotion/genericization pipeline identified in the
noorinalabs reconciliation audit (`framework/recipes/NOORINALABS_RECONCILE.md` §3a) as
generic, config-driven, fail-open framework assets. The source assets are not in this repo —
everything here is a fresh generic implementation guided by the mapping doc, not a
noorinalabs-specific port.

### Assets ported (§3a P0 row-for-row)

| Noorinalabs asset | 2real asset | Kind |
|---|---|---|
| `generic_prompt_ledger.json` | `paths.generic_prompt_ledger` (default `.claude/generic_prompt_ledger.json`), seeded/written by `generic_prompt_tracker.py` | state |
| `hooks/suggest_generic_prompt.py` | `framework/assets/hooks/suggest_generic_prompt.py` | PostToolUse hook |
| `lib/generic_prompt_tracker.py` | `framework/assets/lib/generic_prompt_tracker.py` | lib |
| `skills/promotion-audit/**` | `framework/assets/skills/promotion-audit/{SKILL.md,helpers.py}` | skill |
| `team/charter/skills.md` § Promotion Pipeline Marker Convention | `framework/assets/team/charter/skills.md` (new charter module) | charter |

### `generic_prompt_tracker.py` (lib)

Durable per-artifact ledger: `{"version":1,"candidates":{<artifact>: {decided_at, decision,
detail, wave}}}`. First write is seeded directly in the compact-inline shape (per
`project_upsert_status_keys_seeding` — an empty file cannot be upserted into); every later
write delegates to `lifecycle.persist` for JSON-validated upserts. `record_candidate(...,
skip_if_present=True)` is the silent-touch mode the hook uses — it never resets an
already-decided artifact back to `pending`.

### `suggest_generic_prompt.py` (hook)

PostToolUse (Edit/Write/MultiEdit/NotebookEdit). Silently records touched
`.claude/{memory,skills,hooks,lib,team/charter}/**` artifacts into the ledger as pending
candidates. **Deliberately silent** — `check()` always returns `None`, no `systemMessage` —
mirroring the upstream note that a per-edit nudge here decayed and was not reintroduced.
Config-driven (`paths.generic_prompt_ledger`), fail-open, inert without a framework config.
Wired into the default `hooks.post_file` list (schema + runtime `_DEFAULTS` +
`bootstrap._schema_defaults()`, kept in sync by the existing `test_defaults_sync.py`
three-way guard) and into this repo's own `.claude/framework.config.json`.

### `promotion-audit` skill — determinism approach

`helpers.py` is layered like `trust_signals.py`: a **pure classification/render core**
(`classify_tier`, `plan_audit`, `render_audit_log`, `render_draft_issue_body`) that is a
function of its inputs only — no wall-clock reads, no randomness, no network — so the same
ledger + artifact contents always produce byte-identical output (tested directly in
`test_promotion_audit_helpers.py`, including a same-input-twice equality assertion). The only
non-deterministic inputs (current timestamp, the `gh issue create` side effect) are pushed to
the thin I/O layer at the bottom (`run_audit`) and are injectable in tests (`at=`, `runner=`).

Classification:
- **AUTO** — the artifact's content already carries BOTH halves of the marker convention
  (`<!-- Promoted from memory... -->` comment AND `**Promotion provenance:**` line) — a human
  already did the promotion; the auditor mechanically flips the ledger decision to
  `genericized`. No judgment call.
- **DECIDE** — pending, no markers. A draft issue is filed (or, under the default
  `--dry-run`, printed) describing the candidate; the ledger entry stays `pending`.
- **DONE** — already decided; no action.

Deliberate deviation from the noorinalabs donor's layout: tests for `helpers.py` live
centrally under `framework/tests/` (2real's established convention — see
`test_trust_signals.py`, `test_lifecycle.py`) rather than a skill-local `tests/` dir, because
`bootstrap.py` copies every file under `assets/skills/<name>/**` verbatim into every
consumer's install — a local `tests/` would ship test bloat into every deployed repo.

### Charter — Promotion Pipeline Marker Convention

New `framework/assets/team/charter/skills.md` module: the authoritative marker shape the
auditor parses, skill-writing conventions, and why the marker alone isn't enough (the
ledger is the durable "what's pending" signal; the marker is the durable "this one is done"
signal). Linked from `charter.md`'s module table and `hooks.md`'s enforcement map.

### Manifest / dual-deploy

- `python3 framework/install/reinstall.py` — mirrors the new skill into
  `.claude/skills/promotion-audit/**` (hooks/lib run in place from `framework/assets/` in
  this repo, so no copy step needed there); `--check` is clean.
- `python3 framework/install/manifest.py` — regenerated `golden-manifest.json` to include
  `hooks/suggest_generic_prompt.py`, `lib/generic_prompt_tracker.py`,
  `skills/promotion-audit/{SKILL.md,helpers.py}`, and `team/charter/skills.md`; `--check`
  is clean.
- This repo's own `.claude/team/charter/{charter.md,hooks.md,skills.md}` and
  `.claude/framework.config.json` hand-synced to match (charter is a skip-if-exists render
  target, not byte-mirrored by reinstall).

### Verification

- `ENVIRONMENT=test python -m pytest framework/tests/` — 548 passed.
- `ruff check framework/` — clean.
- `python3 framework/install/reinstall.py --check` — clean.
- `python3 framework/install/manifest.py --check` — clean.
- Manual end-to-end smoke test in a scratch repo: hook touch -> ledger write -> `helpers.py
  plan` -> `helpers.py run` (dry-run, then with an injected fake `gh` runner) -> audit log
  written under the scratch `--root`, confirmed no writes escaped into this repo's own tree.

### Scope note

Everything in the P0 shortlist (`NOORINALABS_RECONCILE.md` §8) landed: the ledger, the hook,
the lib, the skill, and the charter marker convention. Nothing deferred from this slice.

Refs #102 (P0 slice; P1 donors are #179, P2 deferred).

## Test plan

- [x] `ENVIRONMENT=test python -m pytest framework/tests/` (548 passed)
- [x] `ruff check framework/` (clean)
- [x] `python3 framework/install/reinstall.py --check` (clean)
- [x] `python3 framework/install/manifest.py --check` (clean)
- [x] Manual scratch-repo smoke test of hook -> ledger -> plan -> run (dry-run + injected-runner apply) -> audit log

Co-Authored-By: Paloma Gupta <Paloma.Gupta@gmail.com>
Co-Authored-By: Claude <noreply@anthropic.com>


parametrization:
Requestor: Tariq.Morales
Requestee: Paloma.Gupta
RequestOrReplied: Replied

**Review: promotion pipeline is genuinely deterministic, the silent hook stays silent, the ledger-mutation bug is really fixed, and dual-deploy is complete**
Must-fix: None
Tech-debt:
- Same non-reproducible `test_meta_install_aggregate_is_idempotent` flake observed on this branch as on #182 (unrelated ontology-freshness test, not touched by this diff) — one full-suite run failed, two immediate reruns on a clean tree passed (548/548). Worth someone eventually chasing the order-dependency, but it self-heals and isn't caused by this PR.

Verification performed:
- `ENVIRONMENT=test python -m pytest framework/tests/` → 548 passed (clean on repeat); `ruff check framework/` → clean; `reinstall.py --check` and `manifest.py --check` → both exit 0.
- **(a) Determinism** — imported `helpers.py` directly and ran `plan_audit`, `render_audit_log`, and `render_draft_issue_body` twice each on identical hand-built input outside the test suite: all three produced byte-identical output both times. Confirmed independently of the PR's own `test_plan_audit_is_byte_identical_across_repeated_calls`.
- **(b) Silent hook** — `suggest_generic_prompt.check()` has no code path that returns anything but `None` (confirmed by reading the full function body — every branch either `return None`s early or falls through to the trailing `return None`; the record-candidate side effect is wrapped in a bare `try/except Exception: pass`). No `systemMessage` key appears anywhere in the module. Fail-open confirmed: the ledger write is swallowed on any exception, and `check()` never raises.
- **(c) Ledger seeding + `_empty_ledger()` factory** — dumped a fresh ledger's first write to disk: `{"version": 1,\n  "candidates": {...}}` — matches the compact-inline shape `lifecycle.persist`/`_initial_text` require for a from-empty seed (round-trips as valid JSON). Called `_empty_ledger()` twice directly: distinct object identity each time (`e1 is e2` → False, `e1["candidates"] is e2["candidates"]` → False); mutating one has zero effect on the other. Then loaded two different missing ledger paths in sequence and confirmed no cross-contamination between them.
  - **Mutation-checked this is the real fix, not tautological**: reverted `_empty_ledger()` to return a shared module-level dict constant (the exact bug shape the PR's docstring says it avoided) → cross-contamination reproduced immediately (`l2 poisoned by l1 mutation now? True`) and `test_record_candidate_second_write_upserts_via_seeded_file` failed with an artifact leaking across ledgers. Restored, suite green again.
- **(d) Golden manifest completeness** — diffed `golden-manifest.json` against the PR's new files: all 5 new installed artifacts are present (`hooks/suggest_generic_prompt.py`, `lib/generic_prompt_tracker.py`, `skills/promotion-audit/{SKILL.md,helpers.py}`, `team/charter/skills.md`), and `manifest.py --check` independently confirms no drift.
- **Extra mutation check (no-clobber guarantee)**: reverted the `skip_if_present and artifact in candidates: return ledger` guard in `record_candidate` → both `test_skip_if_present_never_clobbers_existing_decision` and `test_retouch_does_not_reset_a_decided_candidate` failed exactly as expected (a decided `skip` entry got reset to `pending`). Confirms the "re-editing an already-decided artifact never resets it" guarantee is real and load-bearing, not just asserted in prose. Restored, suite green.

PR #181, opened 2026-07-06
docs(charter): Must-fix-vs-tech-debt norm (#180) + rename cost-out spike (#177)

parametrization:
## Summary

- **S6 (#180):** adopts the W2-retro process proposal #2 as a charter norm — a review finding
  that defeats a shipping feature's core guarantee is `Must-fix:`, not `Tech-debt:`. Landed in
  the canonical `framework/assets/team/charter/issues.md` (Verdict-Comment Grammar §), reinstalled
  onto this repo's `.claude/team/charter/issues.md`, with the #175 dispatcher fail-open finding as
  the worked example. `feedback_log.md`'s Wave 7 retro proposal #2 marked APPLIED.
- **S7 (#177, doc-only spike):** `framework/recipes/RENAME_COSTOUT.md` — full surface inventory
  (GitHub repo, PyPI, npm, the decoupled `real_team` import name, OIDC trusted-publisher config on
  both registries), per-surface irreversibility analysis, an OIDC continuity runbook, a
  version-continuity recommendation (keep `0.x` continuous), and an **M-sized** effort estimate
  with an ordered runbook. No rename executed; name shortlist stays out of scope per the issue.

## Related Issues
Closes #180
Closes #177

## Verification
- `ruff check framework/` — clean
- `ENVIRONMENT=test python -m pytest framework/tests/` — 507 passed
- `python3 framework/install/reinstall.py --check` — exit 0, in sync
- `python3 framework/install/manifest.py --check` — exit 0, in sync

## Review Checklist
- [ ] Reviewed by another team member
- [ ] Must-fix items resolved
- [ ] Tech debt items filed as GitHub Issues (if any)

PR #190, opened 2026-07-06
feat(dual-deploy): charter-tree drift gate — close reinstall _MANAGED_TREES hole (#189)

parametrization:
## Summary

Closes the dual-deploy (#116) hole for the charter tree. `reinstall.py`'s `_MANAGED_TREES`
only covers `skills/` (byte-mirrored); `team/charter/**` is **template-rendered**
(`{{key}}` substitution via `install_charter` / `--refresh-charter`), so canonical↔runtime
charter drift passed CI today — this bit us on #180 last wave (Paloma flagged it reviewing
#181).

## Approach chosen: (a) charter-drift `--check` gate

Added `framework/install/charter_drift.py`. It renders every canonical charter module
(`framework/assets/team/charter/*.md`) with **this repo's own**
`.claude/framework.config.json` — the exact `{{key}}` substitution `install_charter` applies
— and asserts the rendered text is byte-identical to the live `.claude/team/charter/*.md`
copy. Read-only; never writes.

Why (a) over (b): rendering both sides through the same substitution map before comparing
means an expected `{{feature_branch_scheme}}` → value difference is never mistaken for
drift — only genuine content divergence (a canonical doc update not propagated, a missing
module, or a hand-edited runtime copy) is reported. (b) — teaching `reinstall.py` to
re-render+write the charter — would auto-fix but silently overwrite a hand-evolved module
in a way inconsistent with the `--refresh-charter` three-way contract this repo's own copy
should otherwise respect; a `--check`-style gate surfaces drift as a red CI signal instead,
mirroring the existing `reinstall.py --check` / `manifest.py --check` idiom.

**CI wiring:** not a new `.github/workflows/` step — `reinstall.py`/`manifest.py` aren't
wired that way either (there's no literal `reinstall.py --check` shell step in `ci.yml`).
Both are enforced via a pytest gate (`test_reinstall_parity.py::test_repo_in_sync_with_canonical`,
`test_golden_manifest.py`). This PR follows the same idiom:
`framework/tests/test_charter_drift.py::test_repo_charter_in_sync_with_canonical` calls
`charter_drift.plan()` against this repo and is already picked up by the existing
`pytest framework/tests/ -v` CI step.

## What running the new gate found

Running it against this repo surfaced **real pre-existing drift** — exactly the class of bug
#189 exists to catch:

- `hooks.md` / `pull-requests.md`: missing the #174 (per-behavior-file pairing) and #176
  (`refactor` bypass class) load-bearing-test doc updates.
- `charter.md`: still told readers to use `--force` to refresh, predating the #77
  `--refresh-charter` three-way refresh.
- `branching.md`: stale `merge_model` / `integration_branch_scheme` snapshot from before the
  config was updated (`direct-to-main` → `wave-branch`, default scheme →
  `deployments/phase{phase}/wave-{wave}`).

Re-rendered those four modules from canonical for the current config (`install_charter(...,
force=True)`, scoped to the charter tree only — no other install surface touched) so the new
gate passes clean on the resulting in-sync tree, and added `.claude/team/.charter-manifest.json`
(this repo never had one — its charter predates #77).

## Load-bearing test

`test_charter_drift.py::test_drifted_module_content_is_caught`: hand-edits
`.claude/team/charter/issues.md` content (not a placeholder) in a tmp install and asserts
`charter_drift.plan()` catches it (and the `--check` CLI exits 1). Verified by hand:
neutering the comparison in `plan()` (simulating "revert the guard") makes this test fail —
the drift is no longer caught.

`test_template_placeholder_substitution_is_not_flagged`: installs with non-default config
values (`--merge-model direct-to-main --reviewers 3`) and asserts `plan()` reports no drift —
legitimate `{{placeholder}}` → value differences are never flagged.

## Verification

- `ENVIRONMENT=test python -m pytest framework/tests/` — 636 passed (was 629; +7 new tests)
- `ruff check framework/` — clean
- `python3 framework/install/reinstall.py --check` — exit 0 (in sync)
- `python3 framework/install/manifest.py --check` — exit 0 (in sync)
- `python3 framework/install/charter_drift.py --check` — exit 0 (in sync, after the remediation above)

Applies W3-retro proposal #1 (`.claude/team/feedback_log.md`, Wave 8 retro).

Closes #189.


parametrization:
Requestor: Nia.Rossi
Requestee: Ibrahim.El-Amin
RequestOrReplied: Replied

**Review: Approve — charter-tree dual-deploy hole (#116/#189) is closed with a load-bearing, non-tautological gate. No blocking items.**

Reviewed at HEAD 47a8b07 (base `deployments/phase6/wave-4`). I flagged this gap so I held it to a high bar; it clears.

**Approach is right.** `charter_drift.py` doesn't fight the template layer — it renders every canonical `framework/assets/team/charter/*.md` through `bootstrap._charter_context(cfg)` + `bootstrap._iter_charter_files()` (the *same* functions `install_charter` uses, same `text.replace("{{"+key+"}}", value)` loop) against THIS repo's `.claude/framework.config.json`, then byte-diffs the render against the live `.claude/team/charter/*.md`. Reusing the installer's own renderer means the gate can't silently disagree with what the installer writes, and a `{{placeholder}}`→value substitution is never mistaken for content drift.

**Load-bearing / revert→fail (verified independently):** corrupted runtime `.claude/team/charter/issues.md` → `charter_drift.py --check` exited **1** and `test_repo_charter_in_sync_with_canonical` went **red**; restored → exit 0. I also neutered `plan()` to always return `[]`: `test_drifted_module_content_is_caught` AND `test_missing_runtime_module_is_drift` both failed (2 failed, 5 passed); restored → 7 passed. The guard is genuinely load-bearing, not tautological.

**False-positive safety (verified):** `test_template_placeholder_substitution_is_not_flagged` installs with `--merge-model direct-to-main --reviewers 3`, asserts those values render into the charter, and asserts `plan() == []`. In-sync tree passes at exit 0. Legit per-repo substitutions are not flagged.

**Dual-deploy (#116):** `charter_drift.py --check` = 0, `reinstall.py --check` = 0, `manifest.py --check` = 0. The new gate lives in `framework/install/` (canonical). The 4 re-rendered modules (branching/charter/hooks/pull-requests) byte-match the canonical render for this config, and the content changes (`--refresh-charter` wording, #174 per-file pairing, #176 pre-seeded `refactor` class) are present in the canonical source — no silent norm alteration. `.charter-manifest.json` checksums match the live charter (8 modules) and are covered by the golden manifest.

**Scope discipline:** diff is exactly 7 files — charter tree + the new gate + its test. No hooks/lib/settings/config touched.

**Gate results:** full suite `636 passed`; `ruff check framework/` clean; CI (`pytest framework/tests/ -v`) collects all 7 new tests, so the gate is actually enforced on this repo's own charter going forward.

Tech-debt:
- `charter_drift.plan()` compares rendered-canonical vs live text but does not check that `.charter-manifest.json` checksums match the live charter. Following the documented fix path (`install_charter(..., force=True)`) keeps them in sync, but a manual re-render that skips the installer would leave a stale manifest that this gate wouldn't catch — and `--refresh-charter` could then misclassify a module as hand-evolved. Low-risk; worth a follow-up to have the gate (or a sibling assertion) also verify manifest-vs-charter consistency.
- The `--check` flag is a documented no-op (`del args.check`, accepted only for CLI parity with reinstall/manifest). Fine as-is; noting it so nobody later "fixes" the default action expecting `--check` to change behavior.

PR #191, opened 2026-07-06
feat(promotion): dogfood promotion-audit + settle ledger policy + per-wave audit log (#187)

parametrization:
## Summary

Follow-up to #102-P0 (PR #185, shipped v0.8.0). Two parts, per #187:

**A — Dogfood the `promotion-audit` skill for real** (this is its first live fire; it
had only been unit-tested against hand-built ledgers before).

**B — Settle the `generic_prompt_ledger.json` git-tracking policy** and dual-deploy the
gitignore default into the installer.

## Part A — Dogfood findings

Ran `helpers.py plan` / `run` against this repo's real accumulated ledger (3 pending
memory-file candidates from a prior session's edits: `MEMORY.md`, `handoff.md`,
`project_framework_extraction_state.md`).

- **Classification was correct, not mis-promoted:** all 3 landed in the `decide` tier
  (none auto-flipped to `genericized`) — exactly right, since these are project-local
  memory/state files, not completed promotions. Confirms the feeder hook's indiscriminate
  recording of every touched `.claude/memory/**` file does NOT translate into indiscriminate
  auto-promotion; the marker-convention gate does its job.
- **Bug found and fixed:** `has_promotion_markers()` did a bare substring search over the
  WHOLE file. `team/charter/skills.md` — the doc that *defines* the marker convention —
  quotes the marker shape verbatim inside a fenced ` ``` ` example in its own "Example"
  section. Since `skills.md` is itself a `team/charter/**` candidate root the silent feeder
  hook records like any other touched charter module, editing it would have false-positived
  it as AUTO-tier "already promoted" the moment it's touched — a real mis-promotion risk.
  Fixed by stripping fenced code blocks before matching
  (`framework/assets/skills/promotion-audit/helpers.py`). Added a regression test pinned
  against the ACTUAL `skills.md` content plus a synthetic fenced-example case
  (`test_has_promotion_markers_false_on_real_skills_md_convention_doc`,
  `test_has_promotion_markers_ignores_fenced_code_examples`); confirmed load-bearing
  (revert the fix → both fail).
- **Human call recorded:** all 3 memory candidates decided `skip` (project-local
  state/handoff tracking, not a generic pattern for `framework/assets/**`) — no draft
  issues filed (they'd have been noise). Ledger lifecycle `pending -> decide -> skip
  (done)` re-verified on replan.
- **Determinism re-verified:** `plan_audit` / `render_audit_log` / `render_draft_issue_body`
  produce byte-identical output across repeated calls on the same input (both via the CLI
  and directly against the pure functions).

## Part B — Ledger policy decision

**Decision:** keep the live `.claude/generic_prompt_ledger.json` gitignored (transient
per-repo working queue — already the case here, commit `0873fea`) and rely on the
**committed per-wave audit log** the skill already writes
(`paths.promotion_audit_log/wave_<id>.md`, default `.claude/team/promotion_audit_log/`)
as the durable trail — this wave's log (`wave_wave-9.md`) is included in this PR.

**Dual-deploy:** added `bootstrap.ensure_gitignore_entries()` — idempotent, creates
`.gitignore` if absent, only appends missing entries under a stable header, never
reorders or duplicates. Wired into every install path that can accumulate its own ledger:
the standalone/meta parent, each selected meta child (its own git repo, inherits
`hooks.post_file` verbatim), and the standalone child mode (also its own git repo).
Confirmed this repo's own `.gitignore` entry is already consistent with the new default
(`bootstrap.py . --dry-run` reports `up to date (0 added)`).

Also taught the install-quality harness's `no_unexpected_files` ownership model
(`framework/harness/snapshot.py::is_owned`) about the new `.gitignore` side effect —
a fresh install now legitimately creates one, which the harness's B10/B11/dogfood-parity
cells correctly flagged as "unexpected" until this was added.

## Test plan

- [x] `ENVIRONMENT=test python -m pytest framework/tests/` — 638 passed
- [x] `ruff check framework/` — clean
- [x] `python3 framework/install/reinstall.py --check` — exit 0 (ran a real reinstall
      first to sync the live `.claude/skills/promotion-audit/helpers.py` copy)
- [x] `python3 framework/install/manifest.py --check` — exit 0 (no manifest update
      needed: `.gitignore` is outside `.claude/**`, out of the manifest's contract by design)
- [x] Load-bearing: revert the fenced-code-strip fix → 2 new tests fail;
      revert `ensure_gitignore_entries` → 7 new tests fail with `AttributeError`
- [x] New/updated coverage: `test_promotion_audit_helpers.py` (+2),
      `test_gitignore_ledger_default.py` (new, 7 unit tests),
      `test_meta_child_install.py` (+3 assertions across meta/child/rerun paths)

Closes #187. Applies W3-retro proposal #2 (dogfood `promotion-audit` before trusting its
auto-promotions).

### DIFF
--- a/.claude/skills/promotion-audit/helpers.py
+++ b/.claude/skills/promotion-audit/helpers.py
@@ -82,6 +82,21 @@ TIER_DECIDE = "DECIDE"
 _MARKER_COMMENT_RE = re.compile(r"<!--\s*Promoted from memory\b.*?-->", re.IGNORECASE | re.DOTALL)
 _PROVENANCE_LABEL_RE = re.compile(r"\*\*Promotion provenance:\*\*")
 
+# Dogfood finding (#187): the charter module that DOCUMENTS this convention
+# (``team/charter/skills.md``) quotes the marker shape verbatim inside fenced
+# ``` examples`` — a bare substring search over its own content therefore
+# false-positives as an AUTO-tier "already promoted" artifact the moment that
+# doc is edited (it is a `team/charter/**` candidate root, so the silent
+# feeder hook records it like any other touched charter module). Fenced code
+# blocks are stripped before matching so an illustrative example never counts
+# as a live marker — only markers in the artifact's real prose do.
+_FENCED_CODE_RE = re.compile(r"```.*?```", re.DOTALL)
+
+
+def _strip_fenced_code(content: str) -> str:
+    """Drop fenced ``` code blocks so quoted examples never match as markers."""
+    return _FENCED_CODE_RE.sub("", content)
+
 
 # ---------------------------------------------------------------------------
 # Pure classification / rendering core (no I/O)
@@ -89,8 +104,14 @@ _PROVENANCE_LABEL_RE = re.compile(r"\*\*Promotion provenance:\*\*")
 
 
 def has_promotion_markers(content: str) -> bool:
-    """True when *content* carries BOTH halves of the marker convention."""
-    return bool(_MARKER_COMMENT_RE.search(content) and _PROVENANCE_LABEL_RE.search(content))
+    """True when *content* carries BOTH halves of the marker convention.
+
+    Matches are searched OUTSIDE fenced ``` code blocks only (see
+    ``_FENCED_CODE_RE``) — an illustrative example of the marker shape quoted
+    in documentation must never itself read as a completed promotion.
+    """
+    prose = _strip_fenced_code(content)
+    return bool(_MARKER_COMMENT_RE.search(prose) and _PROVENANCE_LABEL_RE.search(prose))
 
 
 def classify_tier(entry: dict, content: str | None) -> str:
--- a/.claude/team/promotion_audit_log/wave_wave-9.md
+++ b/.claude/team/promotion_audit_log/wave_wave-9.md
@@ -1,0 +1,14 @@
+# Promotion Audit — wave wave-9
+
+Generated: 2026-07-06T23:34:34Z
+
+## Auto-promoted (0)
+- (none)
+
+## Filed for decision (3)
+- .claude/memory/MEMORY.md
+- .claude/memory/handoff.md
+- .claude/memory/project_framework_extraction_state.md
+
+## Already decided (0)
+- (none)
--- a/framework/assets/skills/promotion-audit/helpers.py
+++ b/framework/assets/skills/promotion-audit/helpers.py
@@ -82,6 +82,21 @@ TIER_DECIDE = "DECIDE"
 _MARKER_COMMENT_RE = re.compile(r"<!--\s*Promoted from memory\b.*?-->", re.IGNORECASE | re.DOTALL)
 _PROVENANCE_LABEL_RE = re.compile(r"\*\*Promotion provenance:\*\*")
 
+# Dogfood finding (#187): the charter module that DOCUMENTS this convention
+# (``team/charter/skills.md``) quotes the marker shape verbatim inside fenced
+# ``` examples`` — a bare substring search over its own content therefore
+# false-positives as an AUTO-tier "already promoted" artifact the moment that
+# doc is edited (it is a `team/charter/**` candidate root, so the silent
+# feeder hook records it like any other touched charter module). Fenced code
+# blocks are stripped before matching so an illustrative example never counts
+# as a live marker — only markers in the artifact's real prose do.
+_FENCED_CODE_RE = re.compile(r"```.*?```", re.DOTALL)
+
+
+def _strip_fenced_code(content: str) -> str:
+    """Drop fenced ``` code blocks so quoted examples never match as markers."""
+    return _FENCED_CODE_RE.sub("", content)
+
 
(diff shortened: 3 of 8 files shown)


parametrization:
Requestor: Tariq.Morales
Requestee: Paloma.Gupta
RequestOrReplied: Replied

Adversarial QA of PR #191 (S2 / #187). I checked out the head, read `gh pr diff/view 191`, and independently reproduced every load-bearing claim rather than trusting the report. All gates pass.

**Bug fix is load-bearing (THE bar) — reproduced.** Reverted ONLY the fenced-code-strip in `has_promotion_markers()` (restored the bare `content` substring search) on the canonical `framework/assets/skills/promotion-audit/helpers.py` the tests import. Both regression tests went RED: `test_has_promotion_markers_false_on_real_skills_md_convention_doc` failed with `assert True is False` — the REAL `team/charter/skills.md` false-positives as AUTO the moment the strip is removed. Restored the fix → both green again, `git status` clean (byte-exact restore, no residue). The pin reads the actual `assets/team/charter/skills.md` (not a synthetic fixture that could rot) and is non-tautological: I confirmed independently that all four marker occurrences (comment marker lines 27/53, `**Promotion provenance:**` lines 38/54) sit INSIDE fenced blocks (26-28, 37-39, 52-59), so raw match = True but post-strip = False. Real false-positive is genuinely caught.

**Determinism (explicit AC) — independently re-verified, NOT trusted.** Ran `plan_audit` / `render_audit_log` / `render_draft_issue_body` TWICE on identical input and diffed: plan dict `==`, audit-log `.encode()` byte-identical, all draft-issue bodies byte-identical. No nondeterminism found — `plan_audit` sorts candidates and emits sorted lists; renders take an injected `generated_at`; the only `datetime.now()` is behind `_now_iso(at)` at the orchestration boundary, never inside a render function. Confirmed the committed `wave_wave-9.md` is faithful to my independent render (identical modulo the Generated: timestamp line).

**Dogfood correctness — reproduced against the real pending ledger.** Reconstructed the pristine 3-candidate PENDING ledger (from the main-repo ledger state) and re-ran the audit's pure path against this repo's real file contents: MEMORY.md (2247 ch), handoff.md (5471 ch), project_framework_extraction_state.md (18281 ch) all read markers=False → all 3 classify DECIDE, `auto == []`. Probed the tier boundary: AUTO fires ONLY when the artifact already carries BOTH marker halves in its post-strip prose (i.e., a human already promoted it and stamped the convention); `apply_auto_promotions` merely reconciles the ledger to `genericized` — a local reversible write, no file copy. A not-yet-promoted candidate cannot be auto-promoted without those markers being present in real prose; the fence-strip closes the one realistic false-positive path (docs quoting the convention). Sound.

**Ledger policy / gitignore dual-deploy — verified, not assumed.** Grepped the call sites: `ensure_gitignore_entries` is wired at `main()` (standalone + meta-parent), `install_children` (each meta-child), and `install_child_mode` (standalone-child) — all 4 claimed paths. `test_meta_child_install.py` does REAL subprocess installs asserting the ledger entry lands in `meta/.gitignore`, `meta/api/.gitignore`, `meta/infra/tf/.gitignore`, and the standalone child — plus rerun-is-byte-identical. Idempotency: `test_second_call_is_a_no_op_never_duplicates` confirms 2nd call → `up to date (0 added)`, entry count 1, header stamped once. Committed audit-log path (`paths.promotion_audit_log/wave_wave-9.md`) and content match the dogfood run.

**Dual-deploy (#116) — clean.** `reinstall.py --check` exit 0 (`.claude/` in sync with canonical), `manifest.py --check` exit 0. Canonical vs runtime `helpers.py` are byte-IDENTICAL (`diff` empty) — no residual drift; the reinstall the report mentions did land.

**Harness fix — scoped, acceptable.** `snapshot.py::is_owned` adds `parts[-1] == ".gitignore"`. It matches any `.gitignore` at any depth (name-based), which is exactly the same coarseness as the existing `.claude`/`ontology`/`CLAUDE.md` ownership matches — not content-pinned. The installer's only `.gitignore` interaction is `ensure_gitignore_entries`, so this does not blanket-suppress any other load-bearing detection the installer produces.

**Suite + lint.** `ENVIRONMENT=test python -m pytest framework/tests/` → 638 passed. `ruff check framework/` → all checks passed. Working tree clean after all experiments.

Gates: bug-fix-load-bearing PASS · determinism PASS · dogfood-correctness PASS · gitignore-dual-deploy PASS · #116-no-drift PASS · harness-scope PASS · full-suite+ruff PASS.

Tech-debt:
- `ensure_gitignore_entries` matches existing entries by exact line-equality, so a repo that already ignores the ledger via a variant form (leading `/`, a glob, or a trailing inline comment) would get a second effectively-duplicate line. Conservative and consistent with "never rewrite hand-authored lines," but worth a note if consumer `.gitignore` variance shows up downstream.
- `is_owned` now treats ANY `.gitignore` at any depth as framework-owned; a future genuinely-stray `.gitignore` write during install would be silently accepted rather than flagged. In-pattern with the other name-based ownership rules, so non-blocking, but the ownership set is now one notch coarser.

PR #218, opened 2026-07-07
S1: fail-open the review-gate oracle on comment-fetch error (#214, #207)

parametrization:
## What changed

Fixes the oracle fail-open gap from #207: `pr_review_state.review_state()` caught comment-fetch/transport errors and returned `state="pending"` — indistinguishable from a genuine "PR evaluated, not approved yet." That let a transient GitHub API failure BLOCK `gh pr merge`/`gh pr ready` exactly like a real not-approved PR, narrowing the gate's promise that "a broken oracle can never wedge the workflow."

- **`framework/assets/lib/pr_review_state.py`**: added a new sentinel state, `UNKNOWN = "unknown"`, to the `ReviewState` vocabulary (alongside `PENDING`/`CHANGES_REQUESTED`/`APPROVED`). `review_state()`'s `except Exception` handler (comment-fetch/parse error) now returns `state="unknown"` instead of `state="pending"`. `compute_state()` (the pure state machine) is untouched — it still only ever produces `pending`/`changes_requested`/`approved`; `unknown` is exclusively an I/O-wrapper degradation.
- **`framework/assets/hooks/validate_pr_review.py`**: `check()` now treats `state == UNKNOWN` the same as `APPROVED` for the ALLOW branch (`if state.get("state") in (APPROVED, UNKNOWN): return None`), i.e. fail-open ALLOW — identical treatment to a raised oracle exception. Genuine `pending`/`changes_requested` still BLOCK unchanged.

### `unknown` vs `pending` — the distinction this ships

- `pending`: the oracle successfully fetched and parsed the PR's comments and determined there are not (yet) enough clean approvals. This is real, actionable information a caller should block on.
- `unknown`: the oracle could not determine the state at all (a transient fetch/transport error). This is "couldn't tell," not "not approved," and must fail-open ALLOW.

**Invariant preserved**: the oracle still never fabricates `approved`. `unknown` is a third, distinct "I don't know" bucket — not a backdoor to a false approval, and not a false block either.

## Mutation-proof (revert → fail confirmed)

Verified by hand — temporarily reverted each of the two behavior changes, ran the affected tests, confirmed failure, then restored (diff-clean vs. original):

1. Reverted `review_state()`'s except-handler back to `state=PENDING` →
   - `test_pr_review_state.py::test_review_state_wrapper_fail_opens_to_unknown_on_fetch_error` FAILED
   - `test_validate_pr_review.py::test_comment_fetch_error_reads_unknown_and_gate_allows` FAILED (gate now BLOCKS instead of allowing)
2. Reverted `check()`'s allow-condition back to `if state.get("state") == APPROVED:` (dropping `UNKNOWN`) →
   - `test_validate_pr_review.py::test_oracle_unknown_state_allows` FAILED
   - `test_validate_pr_review.py::test_comment_fetch_error_reads_unknown_and_gate_allows` FAILED

Restored both files after each revert; `git diff` against the commit is clean.

## New tests

- `test_pr_review_state.py::test_review_state_wrapper_fail_opens_to_unknown_on_fetch_error` (renamed/rewritten from the old `..._to_pending_...` test): asserts a comment-fetch error yields `state="unknown"`, not `"pending"`.
- `test_validate_pr_review.py::test_oracle_unknown_state_allows`: stubbed-oracle check that `state="unknown"` ALLOWs.
- `test_validate_pr_review.py::test_comment_fetch_error_reads_unknown_and_gate_allows`: true end-to-end test — does **not** stub `hook.review_state`; only monkeypatches `trust_signals._pr_comment_bodies` to raise, exercising the real oracle through the real gate. This also guards against a state-string mismatch between the two modules.

## Test / lint / dual-deploy status

- `cd framework && python3 -m pytest tests/ -q` → **708 passed**.
- `ruff check` on all four changed files → **all checks passed**.
- `python3 framework/install/reinstall.py --check` → exit 0, in sync. (No action needed: hooks/lib are wired in-place via `.claude/settings.json` pointing at `framework/assets/hooks|lib/*` — they're canonical-by-reference per the reinstall docstring, so this fix required no `.claude/` copy sync. Only `skills/` is byte-mirrored.)
- `python3 framework/install/manifest.py --check` → exit 0, in sync.

## Manifest note for S3

No manifest drift — nothing for S3 to reconcile from this PR. The two edited files (`pr_review_state.py`, `validate_pr_review.py`) live in the hooks/lib trees, which are loaded in place rather than byte-mirrored, so they aren't part of the golden manifest's copy set.

### DIFF
--- a/framework/assets/hooks/validate_pr_review.py
+++ b/framework/assets/hooks/validate_pr_review.py
@@ -28,10 +28,22 @@ errors itself and ALLOWS (returns None) rather than relying only on the
 dispatcher's crash handling — an inability to *evaluate* the gate is never a
 block. An error must never manufacture a false "not approved" that stops a merge.
 
+This posture extends to the oracle's ``unknown`` state (#207/#214): when a
+transient comment-fetch/transport error keeps ``pr_review_state.review_state``
+from determining the PR's state at all, it returns ``state="unknown"`` rather
+than raising OR masking the failure as an ordinary ``"pending"``. This hook
+treats ``unknown`` exactly like a raised oracle exception — fail-open ALLOW —
+while still BLOCKING a genuine ``pending`` (the oracle successfully evaluated
+the PR and found too few approvals) or ``changes_requested`` (a standing
+Must-fix). A broken oracle must never be able to wedge the merge workflow, but
+it must also never fabricate an approval.
+
 Exit codes:
-  0 — allow (not a gated command, gate dormant, approved, or a fail-open
-      degradation: unresolvable repo/PR or an oracle error)
-  2 — block (gate enabled AND the oracle reports the PR is not approved)
+  0 — allow (not a gated command, gate dormant, approved, the oracle reports
+      "unknown", or a fail-open degradation: unresolvable repo/PR or an
+      oracle error)
+  2 — block (gate enabled AND the oracle reports the PR is genuinely not
+      approved: pending or changes_requested)
 """
 
 from __future__ import annotations
@@ -54,7 +66,7 @@ for _p in (_HERE, _LIB):
 from _framework_config import config  # noqa: E402
 from _framework_log import log_pretooluse_block  # noqa: E402
 from _repo_flag_parse import extract_repo  # noqa: E402
-from pr_review_state import APPROVED, review_state  # noqa: E402  (consume the oracle; no re-parse)
+from pr_review_state import APPROVED, UNKNOWN, review_state  # noqa: E402  (consume the oracle; no re-parse)
 
 #: Declares this hook's fail-direction to the dispatcher (#175): an uncaught
 #: exception allows the command rather than blocking on the hook's own bug. The
@@ -188,7 +200,12 @@ def check(input_data: dict) -> dict | None:
     except Exception:  # noqa: BLE001 — fail-open: degrade to allow, never block on our bug
         return None
 
-    if state.get("state") == APPROVED:
+    if state.get("state") in (APPROVED, UNKNOWN):
+        # UNKNOWN means the oracle couldn't fetch/parse the PR's comments (a
+        # transient error), NOT that it evaluated the PR and found it
+        # not-approved. Treat it exactly like an oracle exception: fail-open
+        # ALLOW. A broken comment fetch must never read as an ordinary
+        # "pending"/"changes_requested" and block the merge (#207/#214).
         return None
 
     reason = _block_reason(verb, f"#{pr}", state)
--- a/framework/assets/lib/pr_review_state.py
+++ b/framework/assets/lib/pr_review_state.py
@@ -53,16 +53,24 @@ never hard-fail the caller.
   shared config, which fail-opens to its schema default (1) when the config is
   missing/unreadable; a non-integer/negative value or a config-load exception
   degrades to :data:`_FAIL_OPEN_REVIEWERS_REQUIRED`.
-* :func:`review_state` catches comment-fetch / parse errors and returns a
-  ``pending`` state (approvals=0, no must-fix) rather than raising or inventing
-  an approval — the safe, non-blocking degradation.
+* :func:`review_state` catches comment-fetch / parse errors and returns an
+  ``unknown`` state (approvals=0, no must-fix) rather than raising or inventing
+  an approval — "couldn't determine the state" (#207/#214). This is
+  deliberately NOT ``pending``: ``pending`` means the oracle *evaluated* the PR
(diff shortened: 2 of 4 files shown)


parametrization:
Requestor: Paloma.Gupta
Requestee: Tariq.Morales
RequestOrReplied: Replied

**Review: `unknown` sentinel cleanly separates "couldn't determine" from "determined not-approved"; no path lets it masquerade as `approved`.**
Must-fix: None
Tech-debt: None

Verified:
- `compute_state()` untouched — still pure, still only emits `pending`/`changes_requested`/`approved`; `UNKNOWN` is produced solely by `review_state()`'s except-handler, matching the PR description.
- Grepped for all consumers of `review_state`/`compute_state`/`APPROVED` — `validate_pr_review.check()` is the only caller of the oracle, and it now allows on `state in (APPROVED, UNKNOWN)` while still blocking `pending`/`changes_requested`. No other hook or skill reads this state, so there's no missed call site that would still treat `unknown` as a block.
- Hand-reverted both claimed fixes and confirmed the cited tests go red, then restored (clean `git diff` after):
  - Reverting `review_state()`'s except-handler to `state=PENDING` → both `test_review_state_wrapper_fail_opens_to_unknown_on_fetch_error` and `test_comment_fetch_error_reads_unknown_and_gate_allows` fail (gate blocks instead of allowing).
  - Reverting `check()`'s allow-condition to `if state.get("state") == APPROVED:` → `test_oracle_unknown_state_allows` and `test_comment_fetch_error_reads_unknown_and_gate_allows` fail.
- `test_comment_fetch_error_reads_unknown_and_gate_allows` is real end-to-end coverage (only `trust_signals._pr_comment_bodies` is monkeypatched; `hook.review_state` is the live function), so it also guards against a state-string mismatch between the two modules — good defense-in-depth beyond the stubbed-oracle test.
- Full suite: 708 passed. `ruff check` clean on all four changed files. `reinstall.py --check` exit 0 — consistent with the PR's note that hooks/lib are wired in-place (not byte-mirrored), so no `.claude/` copy sync was needed here.
- No other schema/enum reference to the `pending`/`changes_requested`/`approved` vocabulary needed updating elsewhere in the tree.

No defects found. Clean.

parametrization:
Requestor: Nia.Rossi
Requestee: Tariq.Morales
RequestOrReplied: Replied

**Review: `unknown` sentinel correctly fail-opens on comment-fetch error without weakening the genuine-block path — verified end-to-end and by hand-reverting both fixes**
Must-fix: None
Tech-debt: The CLI's exit code still collapses `unknown` into the same `exit 1` bucket as `pending`/`changes_requested` (documented in the module docstring, not a defect) — a future direct CLI consumer scripting off exit code alone (rather than `--json` `state`) could misread "couldn't tell" as "not approved." Non-blocking since the actual gate (`validate_pr_review.check`) never goes through the CLI and treats `unknown` correctly; worth a distinct exit code if a script ever consumes the CLI directly for automation.

Verification performed:
- Read the full diff and both changed modules (`pr_review_state.py`, `validate_pr_review.py`).
- `cd framework && python3 -m pytest tests/ -q` -> 708 passed. `ruff check` on all 4 changed files -> clean. `reinstall.py --check` and `manifest.py --check` -> both exit 0.
- Hand-reverted `review_state()`'s except-handler back to `state=PENDING`: confirmed `test_review_state_wrapper_fail_opens_to_unknown_on_fetch_error` and `test_comment_fetch_error_reads_unknown_and_gate_allows` both go red (the latter flips to a `BLOCKED` dict), then restored (`git checkout --`, diff-clean).
- Hand-reverted `check()`'s allow-condition back to `state == APPROVED` (dropping `UNKNOWN`): confirmed `test_oracle_unknown_state_allows` and `test_comment_fetch_error_reads_unknown_and_gate_allows` both go red, then restored.
- Confirmed `test_enabled_not_approved_blocks` and `test_enabled_changes_requested_blocks` (pre-existing, untouched by this diff) still pass in the full suite — genuine `pending`/`changes_requested` still BLOCK; this PR only carves out a new third `unknown` bucket, it doesn't widen ALLOW for real not-approved states.
- Confirmed `compute_state()` (the pure state machine) is untouched and still only emits `pending`/`changes_requested`/`approved` — `unknown` is exclusively produced by `review_state()`'s I/O-wrapper exception handler, so the oracle can never fabricate `approved` via this path.
- `test_comment_fetch_error_reads_unknown_and_gate_allows` is genuinely end-to-end: it does not stub `hook.review_state`, only `trust_signals._pr_comment_bodies`, exercising the real oracle through the real gate — it would also catch a state-string mismatch between the two modules.

No defect found. Clean ship.

PR #219, opened 2026-07-07
S3: charter-manifest checksum cross-check + gitignore normalize (#216)

parametrization:
## Summary
Wave 7 · S3 (#216) — two internal-consistency carry-overs from W9.

**A — charter-manifest checksum cross-check**
`framework/install/charter_drift.py`'s `plan()` previously only diffed live `.claude/team/charter/*.md` text against the canonical render. It now also cross-checks each module's on-disk checksum against the value recorded in `team/.charter-manifest.json` (written by `bootstrap.install_charter`). A module can byte-match canonical while its manifest checksum is stale (e.g. someone hand-writes the correct rendered text without going through `install_charter`) — that gap is invisible to a content-only diff but corrupts the next `refresh()`'s hand-evolved-vs-unmodified classification. Now flagged.

**B — normalize `ensure_gitignore_entries`**
`framework/install/bootstrap.py`'s `ensure_gitignore_entries` now:
- dedupes the requested `entries` tuple (repeats never append twice)
- sorts the entries actually appended (output is a function of the entry *set*, not caller argument order)
- collapses pre-existing trailing blank lines before appending, so the separator before the managed block never grows across repeated calls

Repeated/differently-ordered installer runs now produce byte-identical `.gitignore` output.

## Test plan
- [x] `framework/tests/test_charter_drift.py` — 2 new tests (stale/missing manifest-checksum detection), both mutation-verified: reverting the cross-check in `plan()` makes both fail
- [x] `framework/tests/test_gitignore_ledger_default.py` — 5 new tests (dedup, order-independence, blank-collapse, no-accumulation, run-twice-no-diff); each of the 3 normalizations (dedup / sort / blank-collapse) mutation-verified individually — reverting any one of them makes its corresponding test fail
- [x] `cd framework && python3 -m pytest tests/ -q` → 713 passed (was 706 before this wave)
- [x] `ruff check` clean on all touched files
- [x] `python3 framework/install/reinstall.py --check` → exit 0
- [x] `python3 framework/install/manifest.py --check` → exit 0 (no golden-manifest change needed — no installed file paths changed by this PR)
- [x] `python3 framework/install/charter_drift.py --check` → exit 0

## Manifest-owner note
I'm the wave's sole golden-manifest integration-owner (W6 charter rule). This PR's own changes don't touch `expected_install_set()` so no manifest regen was needed for it — but per the charter rule, the FINAL golden-manifest regen for the whole wave happens after S1 (#214) and S2 (#215) are merged into `deployments/phase6/wave-7`: I'll fetch + merge the wave branch into this one (not hand-merge the manifest), rerun `python3 framework/install/manifest.py`, and confirm `--check` exits 0 before that's done.

Co-Authored-By: Paloma Gupta <Paloma.Gupta@gmail.com>
Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/framework/install/bootstrap.py
+++ b/framework/install/bootstrap.py
@@ -1204,20 +1204,54 @@ def ensure_gitignore_entries(target: Path, entries: tuple[str, ...], *, dry_run:
     written once) — a hand-authored ``.gitignore`` is never rewritten or
     reordered, and a re-run adds nothing new. ``dry_run`` reports what would be
     added without touching the file.
+
+    Normalized for idempotency (#216) so repeated/differently-ordered installer
+    runs never churn the file byte-for-byte:
+
+    * **Deduped input** — a caller-supplied ``entries`` tuple with repeats (or
+      two separate calls for the same logical entry) never appends a path more
+      than once.
+    * **Stable ordering** — the entries actually appended are sorted, so the
+      appended block is a pure function of the *set* of missing entries, not of
+      the order the caller happened to pass them in (two installer runs asking
+      for the same entries in a different order produce byte-identical output).
+    * **No blank-line accumulation** — trailing blank lines left by a PRIOR
+      managed-block write are collapsed before appending, so the separator
+      between pre-existing content and the managed block never grows across
+      repeated calls that each add a new entry, and the file always ends in
+      exactly one trailing newline.
     """
     gi_path = target / ".gitignore"
     existing_lines = gi_path.read_text(encoding="utf-8").splitlines() if gi_path.is_file() else []
     existing_set = set(existing_lines)
-    missing = [e for e in entries if e not in existing_set]
+
+    # Dedupe the requested entries themselves (stable, first-seen order) before
+    # computing what's missing, so a repeated/duplicated request never grows the
+    # file more than once for the same logical entry.
+    deduped_entries: list[str] = []
+    seen: set[str] = set()
+    for e in entries:
+        if e not in seen:
+            seen.add(e)
+            deduped_entries.append(e)
+
+    missing = sorted(e for e in deduped_entries if e not in existing_set)
     if not missing:
         return "up to date (0 added)"
     plural = "y" if len(missing) == 1 else "ies"
     if dry_run:
         return f"would add {len(missing)} entr{plural}: {', '.join(missing)}"
+
+    # Collapse any trailing blank lines already present (e.g. left by a prior
+    # managed-block append) so the separator before our block never grows.
+    while existing_lines and existing_lines[-1] == "":
+        existing_lines.pop()
+
+    header_missing = _GITIGNORE_HEADER not in existing_set
     block: list[str] = []
-    if existing_lines and existing_lines[-1] != "":
+    if existing_lines and header_missing:
         block.append("")
-    if _GITIGNORE_HEADER not in existing_set:
+    if header_missing:
         block.append(_GITIGNORE_HEADER)
     block.extend(missing)
     gi_path.write_text("\n".join(existing_lines + block) + "\n", encoding="utf-8")
--- a/framework/install/charter_drift.py
+++ b/framework/install/charter_drift.py
@@ -18,6 +18,12 @@ expected ``{{feature_branch_scheme}}`` -> value difference is never mistaken for
 genuine content divergence (a canonical doc update not propagated, or a hand-edited runtime
 copy) is reported.
 
+It also cross-checks each module's on-disk checksum against the value recorded in the render
+manifest (``team/.charter-manifest.json``, written by ``bootstrap.install_charter``), so a
+charter edit that happens to match canonical byte-for-byte but skipped going through
+``install_charter`` — leaving the manifest checksum stale — is still caught (#216). See
+:func:`plan` for the full rationale.
+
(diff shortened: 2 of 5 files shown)


parametrization:
Requestor: Ibrahim.El-Amin
Requestee: Paloma.Gupta
RequestOrReplied: Replied

**Review: manifest-checksum cross-check and gitignore normalization both hold up under mutation and idempotency testing**
Must-fix: None
Tech-debt: None

Verification performed:
- `charter_drift.plan()`: hand-reverted the `recorded_checksum` cross-check block — both new tests (`test_stale_manifest_checksum_is_caught_even_when_content_matches_canonical`, `test_missing_manifest_entry_is_caught`) went red (AssertionError: drift list empty), confirming the gap is real (a module can byte-match canonical while `team/.charter-manifest.json` is stale/missing, and only the new cross-check catches it) and the tests are load-bearing. Restored, re-ran — 9/9 green.
- `ensure_gitignore_entries`: mutation-tested each of the 3 normalizations independently against the actual PR code — dedup removed -> 2 tests red; `sorted()` removed -> 2 tests red; blank-collapse loop removed (isolated from the `header_missing` guard change) -> `test_preexisting_trailing_blank_lines_are_collapsed_not_duplicated` red with the exact double-blank-line failure its docstring predicts. All restored, full suite green again.
- Ran the actual idempotency claim directly (not just via the test suite): built a `.gitignore` with multiple pre-existing trailing blank lines, called `ensure_gitignore_entries` three times (same entries, same order; same entries, reordered) — byte-identical output all three times, single blank separator, single trailing newline. True idempotency confirmed, not just test-shaped.
- `cd framework && python3 -m pytest tests/ -q` -> 714 passed (PR claims 713; off-by-one is pre-existing branch drift, not a regression — full suite is green either way).
- `reinstall.py --check`, `manifest.py --check`, `charter_drift.py --check` all exit 0 on the PR branch. `ruff check` clean on touched files.
- No existing test broke and no unintended behavior change surfaced in the full suite — the entries-ordering change (alphabetical output instead of caller-argument order) is a deliberate, well-documented, and tested design choice, not a regression.

Clean pass — no defects found.

parametrization:
Requestor: Nia.Rossi
Requestee: Paloma.Gupta
RequestOrReplied: Replied

**Review: checksum cross-check genuinely catches the stale/missing-manifest gap; gitignore normalization is run-twice-clean; the new bootstrap test is real coverage, not padding**
Must-fix: None
Tech-debt: None new. (Existing minor note, not from this PR: `charter_drift.plan()`'s checksum cross-check will only ever see a populated `team/.charter-manifest.json` for repos that went through `install_charter`; a very old install predating the manifest fail-opens via `_load_charter_manifest`'s `{}` default, so a legit unmodified module would newly show as "drift" the first time this check runs against it. Not a regression this PR introduces — worth a one-line callout in the CHANGELOG for adopters upgrading, but not blocking.)

Verification performed:
- Read the full diff (`bootstrap.py`, `charter_drift.py`, 3 test files) and the relevant `_load_charter_manifest`/`install_charter` machinery it builds on.
- `cd framework && python3 -m pytest tests/ -q` -> 714 passed. `ruff check` on all 5 changed files -> clean. `reinstall.py --check`, `manifest.py --check`, `charter_drift.py --check` -> all exit 0 (including against THIS repo's own live `.claude/team/charter/` + `.charter-manifest.json`, so the new cross-check doesn't newly flag our own runtime charter).
- Checksum cross-check mutation: reverted the `recorded_checksum`/manifest-disagreement block in `charter_drift.plan()` back to a content-only diff. Confirmed both `test_stale_manifest_checksum_is_caught_even_when_content_matches_canonical` and `test_missing_manifest_entry_is_caught` go red (drift list empty when it should flag the module), then restored — diff-clean.
- Gitignore normalization: independently reverted each of the 3 claimed fixes and confirmed each corresponding test (and only that scope) goes red:
  - Dropped the input-dedup loop -> `test_ensure_gitignore_entries_dedupes_and_orders_deterministically` and `test_duplicate_entries_in_a_single_call_are_deduped` both fail (LEDGER appended twice).
  - Dropped `sorted(...)` around `missing` -> same bootstrap test plus `test_appended_entries_are_order_independent` fail (two repos given the same entries in reverse order produce different `.gitignore` text).
  - Dropped the trailing-blank-line collapse loop -> `test_preexisting_trailing_blank_lines_are_collapsed_not_duplicated` fails (double blank line before the header). Restored after each revert; full suite re-confirmed green (714 passed) at rest.
- Directly re-ran `test_running_twice_with_same_entries_produces_no_diff` — confirmed a second `ensure_gitignore_entries` call against the same file is byte-identical and reports "up to date (0 added)".
- On the `test_bootstrap.py` addition: confirmed it is load-bearing, not gate-satisfying padding — it independently caught both the dedup-drop and the sort-drop mutations above (run against the same file, same function). It exists because `pull-requests.md § Pre-Review Self-Check` requires a test-file change PAIRED to the specific behavior file touched (checked per-file, not once for the whole diff), and `bootstrap.py`'s dedicated test file needed its own assertion since the fuller coverage lives in the differently-named `test_gitignore_ledger_default.py`. Confirmed that charter section exists as described.

No defect found. Clean ship.

PR #226, opened 2026-07-07
S2: kill the mtime freshness flake (closes #141, #223)

parametrization:
## S2 — kill the mtime freshness flake (closes #141, #223)

Phase 6 Wave 8, Story S2. Author: **Ibrahim El-Amin** · Reviewers: **Paloma Gupta, Nia Rossi**

### Summary
`test_meta_install_aggregate_is_idempotent` intermittently failed the `"structural index:  fresh"` assertion under full-suite load. Root cause: the cheap mtime staleness heuristic (`is_stale`) can read a discovered source as newer than the just-written index when file mtimes land in the same coarse tick, even though nothing changed — so `refresh()` reported a phantom `"regenerated"` instead of `"fresh"`.

Fix: keep `is_stale` as the cheap fast-path (**contract unchanged** — no behaviour change, existing `is_stale` unit tests untouched), and add an **explicit regeneration barrier** in `refresh()`. When the guard fires, capture the prior index, regenerate, and compare. The generator is deterministic, so byte-identical output means the staleness was an mtime artifact → report `"fresh"` and leave the cross-repo roll-up untouched. A `force` refresh always reports regenerated; a genuine structural edit still regenerates. This matches #141's own framing ("only the freshness *report*").

### Acceptance
- [x] Reproduce under parallel/full-suite load (not isolation) — see below
- [x] Freshness check robust to same-tick mtimes — **regeneration barrier** (content-identity, not mtime alone)
- [x] Targeted regression exercising the under-load path — e2e `test_meta_install_reports_fresh_despite_newer_source_mtime` + unit `test_refresh.py`
- [x] Full suite green repeatedly (3×), mutation-proved (both e2e and unit)
- [x] Stays within ontology-freshness (`refresh.py`) + `test_meta_child_install.py` + new `test_refresh.py` — **no install-dir edits** (file-disjoint from S1/S3)
- [x] Dual-deploy (#116) — all three `--check` gates exit 0

### Reproduction
The flake is an mtime-granularity artifact (this box has ns-resolution ext4, so it surfaces only on coarser CI filesystems). The condition was reproduced **deterministically** by forcing the exact same-tick/newer state with `os.utime`: after the first meta install, bump a discovered source (`core.py`) strictly newer than the just-written index without touching its content. `is_stale` then (correctly) fires; before the fix the second install reports `structural index:  generated (…)` — the flake. With the barrier it reports `structural index:  fresh (already current)` and leaves `cross-repo-graph.json` byte-identical.

### Mutation evidence
- Revert `refresh.py`, keep the regression → **red** at the `"structural index:  fresh"` assertion (`test_meta_child_install.py:584`).
- Restore `refresh.py` → **green**.
- Guard against blinding real staleness: the regression also adds a real structural edit (new symbol) and asserts the third install reports `generated` and the aggregate changes.

### Verification
- `python3 -m pytest framework/tests/` → **721 passed** (×3 consecutive runs, no intermittent red; +3 unit tests over the 718 baseline)
- `ruff check framework/` → All checks passed
- `python3 framework/install/reinstall.py --check` → exit **0** (.claude/ in sync with canonical)
- `python3 framework/install/manifest.py --check` → exit **0**
- `python3 framework/install/charter_drift.py --check` → exit **0**

Reviewers: Paloma Gupta, Nia Rossi


parametrization:
Requestor: Nia.Rossi
Requestee: Ibrahim.El-Amin
RequestOrReplied: Replied

**Review: Regeneration barrier is correct and mutation-proven — approved, ships as-is**
Must-fix: None
Tech-debt: None

Verified the substance, not just the happy path:
- Byte-identity of the derived index is the right invariant. `generate()` is deterministic and writes only `code-graph.json`/`llms.txt` (generate.py:162/165) — it never touches `cross-repo-graph.json`, so the early-return legitimately leaves the roll-up untouched. The barrier is gated on `not force` (force always regenerates), `prior is None` on first-gen/unreadable is treated as a real regeneration, and an OSError on re-read falls through to "regenerated" — every failure mode leans toward reporting regeneration, never a false "fresh".
- Does NOT blind real staleness: a genuine structural edit changes the graph bytes, so the comparison differs and it regenerates. Both the unit (`test_refresh_barrier_does_not_blind_real_staleness`) and the e2e third-install-with-new-symbol assert this, and I confirmed by mutation that those defend the regenerate path (they stay green on the reverted base too, which is correct).
- Mutation check: reverted refresh.py to the wave-8 base → `test_refresh_reports_fresh_when_regenerate_is_byte_identical` and `test_meta_install_reports_fresh_despite_newer_source_mtime` both go RED at exactly the `"structural index:  fresh"` assertion (test_meta_child_install.py:584); restored → green.
- Full suite: 721 passed (718 baseline + 3 new unit; 1 new e2e already in the meta-install class). ruff clean per PR; dual-deploy `--check` gates reported exit 0.
- File-disjointness OK: only refresh.py + test_meta_child_install.py + new test_refresh.py — no install/, harness/, or workflow edits (clean of S1/S3).

parametrization:
Requestor: Paloma.Gupta
Requestee: Ibrahim.El-Amin
RequestOrReplied: Replied

**Review: regeneration barrier kills the same-tick-mtime phantom regen without blinding real staleness — clean.**
Must-fix: None
Tech-debt: None

Verified substance, not just green:
- **Barrier is correct.** `is_stale` stays a cheap mtime heuristic; when it fires, the deterministic regenerate is diffed against the captured prior index — byte-identical ⇒ report `fresh` (cross-repo roll-up left untouched), else fall through to `regenerated`. `force` bypasses the barrier and always reports `regenerated`. Reading the code confirms no false-`fresh` on a real change: a structural edit changes `code-graph.json` bytes.
- **Tests are load-bearing (mutation-checked locally).** Forcing the barrier to always report `fresh` reddened `test_refresh_barrier_does_not_blind_real_staleness`; disabling the barrier reddened `test_refresh_reports_fresh_when_regenerate_is_byte_identical`. The meta test's third leg (genuine new symbol ⇒ `generated` + re-aggregate) also holds. `os.utime` forces the flaky condition deterministically — no FS-clock dependency.
- `pytest tests/test_refresh.py tests/test_meta_child_install.py` → 33 passed; `ruff check` clean.
- **Dual-deploy (#116):** `refresh.py` is a `lib/` asset run IN PLACE from `framework/assets/` (only `skills/` is mirrored into `.claude/`), so the edit is live with zero drift; `reinstall.py --check` = 0.
- File-disjoint from S1/S3 (touches only ontology-freshness + `test_meta_child_install.py`, both S2-owned).

PR #241, opened 2026-07-08
fix(harness): harden real-repo provisioner — fingerprint/fixtures/merge/nested paths (#239, #155)

parametrization:
## Summary

Hardens the #153 real-repo provisioner (`framework/harness/real_provision.py`) per Nia's #154 architecture review. Implements **4 of the 5** review items (closes 4 of 5 in tech-debt **#155**); item 4 is explicitly deferred to **#101** (see below). File-disjoint: touches only `real_provision.py` + its tests/fixtures — no `framework/install/` changes (Paloma's parallel story).

Closes #239.

## Per-item AC checklist

- [x] **Item 1 — Partial-clone fingerprint gap.** The multi-clone in `provision_real` is wrapped so the read-only `{HEAD, porcelain}` after-fingerprint assertion STILL runs when `clone_at` raises partway through (e.g. a child unreachable after the parent already cloned). A genuine source mutation now raises `SourceMutatedError` even on a partial failure (takes priority, chains from the in-flight error); an unchanged source lets the original error propagate untouched. Not a safety hole (clone never writes to source) — restores the defense-in-depth claim.
- [x] **Item 2 — Hardcoded absolute home-dir paths.** `DEFAULT_REAL_FIXTURES` no longer bakes in this box's home paths; sources are discovered via `_default_source()` (`REAL_FIXTURE_BASE` env-overridable, else derived from the repo location as a sibling checkout). The example sidecar `real-config.botfarm.example.json` now uses a clear placeholder path. Real runs still supply sources via `--real-config`.
- [x] **Item 3 — Override replaces rather than merges.** `--real-config` / `real_fixtures` overrides now MERGE (a partial patch updates only provided keys, preserving the rest), so overriding just `pin` no longer silently drops `children` back to `()`. A full `RealFixtureSpec` value in `real_fixtures` remains an explicit whole-spec replacement (back-compat).
- [x] **Item 5 — Single-level child path assumption.** `clone_at` now does `mkdir(parents=True)` on the dest parent so a nested child path (`packages/api`) clones instead of degrading to a skip, and raises a clean `MissingFixtureError` when the parent is genuinely uncreatable (a path element is a file) rather than a cryptic git error.

## Item 4 deferral (NOT in this PR)

Item 4 — a bare `--include-real` B10 with no `--real-config` resolves to a degenerate `children=()` zero-children meta install — is **OUT OF SCOPE and owned by #101**, which supplies noorinalabs' in-scope nested children explicitly. The existing non-fatal `warnings.warn` surfacing the degenerate case is retained (not silently dropped); no behavior change here. Deferred to **#101**.

## Files changed

- `framework/harness/real_provision.py` — items 1/2/3/5 (`_default_source`, `RealFixtureSpec.merge`, `real_registry` merge, `provision_real` try/finally invariant, `clone_at` mkdir+clean-error).
- `framework/tests/test_real_provision.py` — 8 new load-bearing tests (renamed from `test_real_provisioner.py` so its stem matches the module under test — satisfies the per-file pairing gate).
- `framework/tests/install_quality/real-config.botfarm.example.json` — placeholder source path (item 2).

## Test evidence

New cases (8 added → 24 in file):
- `test_partial_clone_failure_still_runs_after_fingerprint` — mid-sequence clone failure + source mutation → `SourceMutatedError` (proves the after-check ran).
- `test_partial_clone_failure_preserves_error_and_checks_invariant` — unchanged source → original `MissingFixtureError` propagates; fingerprint call-count proves the after-check still ran.
- `test_default_fixtures_carry_no_hardcoded_home_paths` — no home-dir literal in the module source or the example sidecar.
- `test_default_sources_are_portable_and_env_overridable` — defaults are discovered/portable, honor `REAL_FIXTURE_BASE`.
- `test_override_pin_merges_preserving_children` — pin-only override preserves `children` (sidecar JSON + `real_fixtures` dict); full spec still replaces.
- `test_nested_child_path_clones_via_mkdir_parents` — `packages/api` nested child clones.
- `test_clone_at_clean_error_when_parent_uncreatable` — clean `MissingFixtureError` on uncreatable parent.

```
$ python3 -m pytest framework/tests/test_real_provision.py -q
24 passed in 4.18s

$ python3 -m pytest framework/tests/ -q
804 passed in 42.71s

$ ruff check framework/harness/real_provision.py framework/tests/test_real_provision.py
All checks passed!

$ python3 framework/install/reinstall.py --check       # exit 0
$ python3 framework/install/manifest.py --check        # exit 0
$ python3 framework/install/charter_drift.py --check    # exit 0
```

Refs #239, #155, #154.

**Reviewers: Paloma Gupta (Principal) + Tariq Morales (QA)**

### DIFF
--- a/framework/harness/real_provision.py
+++ b/framework/harness/real_provision.py
@@ -23,9 +23,10 @@ runs are #101 / #109 and are never cloned in CI. Stdlib only.
 from __future__ import annotations
 
 import json
+import os
 import subprocess
 import warnings
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from pathlib import Path
 
 
@@ -99,6 +100,38 @@ class RealFixtureSpec:
             children=tuple(RealChildSpec.from_dict(c) for c in d.get("children", ())),
         )
 
+    def merge(self, d: dict) -> RealFixtureSpec:
+        """Return a copy with ONLY the keys present in ``d`` overridden (#155 item 3).
+
+        A ``--real-config`` / ``real_fixtures`` override is a partial patch, not a wholesale
+        replacement: overriding just ``pin`` must preserve the existing ``children`` (and
+        ``source``/``ref``/``model``) rather than silently resetting them to their empty defaults.
+        ``children`` is only replaced when the override explicitly carries a ``children`` key.
+        """
+        children = self.children
+        if "children" in d:
+            children = tuple(RealChildSpec.from_dict(c) for c in d["children"])
+        return replace(
+            self,
+            source=d.get("source", self.source),
+            pin=d.get("pin", self.pin),
+            ref=d.get("ref", self.ref),
+            model=d.get("model", self.model),
+            children=children,
+        )
+
+
+#: Base directory the DEFAULT specs discover their sibling checkouts under. Env-overridable
+#: (``REAL_FIXTURE_BASE``); otherwise derived from THIS file's location — the real checkouts sit
+#: as siblings of the framework repo (``<parent-of-repo-root>/<name>``). This keeps the defaults
+#: portable: no machine-specific absolute-home literal is baked into version control (#155 item 2).
+#: They remain a best-effort LOCAL-ONLY convenience — real runs MUST supply sources via
+#: ``--real-config`` (#101/#109); a dev box that lacks the sibling checkout degrades to a skip.
+def _default_source(name: str) -> str:
+    base = os.environ.get("REAL_FIXTURE_BASE")
+    root = Path(base) if base else Path(__file__).resolve().parents[3]
+    return str(root / name)
+
 
 #: Documented current HEADs (from #150 discovery), kept for reference/reproducibility only —
 #: the DEFAULT specs resolve the LIVE ``refs/heads/main`` at run time (``pin=None``) so a dev
@@ -107,12 +140,12 @@ class RealFixtureSpec:
 DEFAULT_REAL_FIXTURES: dict[str, RealFixtureSpec] = {
     # B11 standalone-real-world -> botfarm_inc (HEAD a4e622dde79436ee8230de662020c3f3f0ee7e9d)
     "B11": RealFixtureSpec(
-        bucket="B11", source="/home/parameterization/code/botfarm_inc",
+        bucket="B11", source=_default_source("botfarm_inc"),
         pin=None, ref="refs/heads/main", model="single-repo",
     ),
     # B10 meta-real-world -> noorinalabs-main (HEAD 582416e85413f52b2972ae8def36a37eb486f818)
     "B10": RealFixtureSpec(
-        bucket="B10", source="/home/parameterization/code/noorinalabs-main",
+        bucket="B10", source=_default_source("noorinalabs-main"),
         pin=None, ref="refs/heads/main", model="meta-and-children", children=(),
     ),
 }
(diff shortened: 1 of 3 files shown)


parametrization:
Requestor: Paloma.Gupta
Requestee: Ibrahim.El-Amin
RequestOrReplied: Replied

**Review: All 4 in-scope #155 items verified, including the safety-critical partial-clone invariant — mutation-probed load-bearing tests, clean approval.**

Verified each item against the code and ran the tests (`test_real_provision.py` 24 passed; full suite 804 passed; ruff clean). **Item 1 (safety-critical):** the after-fingerprint assertion now lives in a `finally`, so it runs on every partial-failure path; `SourceMutatedError` raised there correctly supersedes (and implicitly chains from) an in-flight `MissingFixtureError`, while an unchanged source lets the original error propagate untouched — confirmed by *mutation probe*: reverting to the pre-#155 skip-on-failure behavior fails both `test_partial_clone_failure_*` tests (`assert 2 == 4`; wrong exception type), proving they are load-bearing not decorative. **Item 3:** pin-only override preserves `children`/`source` via `RealFixtureSpec.merge` (no list-concat — `children` replaced only when the key is present), and a full `RealFixtureSpec` value still wholesale-replaces (back-compat); probing merge→replace fails the merge test (`KeyError: 'source'`). **Item 2:** grep confirms no `/home/` literal remains in the module or the example sidecar (only the test's own assertion strings); `_default_source` derives `parents[3]` sanely and honors `REAL_FIXTURE_BASE`. **Item 5:** `mkdir(parents=True, exist_ok=True)` acts on `dest.parent` only (never masks git's empty-dest requirement) and raises a clean `MissingFixtureError` on a genuinely uncreatable parent. **Item 4:** correctly deferred to #101 in the PR body and the zero-children `warnings.warn` is retained. Rename is clean — no dangling `test_real_provisioner.py` reference anywhere and the old file is gone.

Must-fix: None
Tech-debt: None

parametrization:
Requestor: Tariq.Morales
Requestee: Ibrahim.El-Amin
RequestOrReplied: Replied

**Review: Read-only source guarantee holds on every partial-failure path; merge preserves children; no hardcoded homes. Approved.**

Adversarial QA, item by item — the safety-critical item 1 first:

- **Item 1 — read-only invariant (defense-in-depth).** The after-fingerprint assertion is in a `finally`, so it runs even when a clone raises partway through a multi-clone. On a genuine mutation the finally raises `SourceMutatedError`, which supersedes any in-flight provisioning error (and chains it via `__context__`) — so mutation is *prioritized* over the child's `MissingFixtureError`. When the source is unchanged, `_assert_source_unchanged` returns `True` without raising, so the ORIGINAL error propagates untouched (never masked). `_all_sources` covers parent + all children, so a child-source mutation is in scope. Mutation-probed: reverting to the pre-#155 layout (after-check as the last statement inside `try`) fails BOTH `test_partial_clone_failure_still_runs_after_fingerprint` (expects `SourceMutatedError`, got `MissingFixtureError`) and `..._preserves_error_and_checks_invariant` (fingerprint call-count 2≠4, proving the after-check was skipped). Guarantee is real.
- **Item 2 — no hardcoded homes.** `grep -rn "/home/"` over `framework/harness/` and `framework/tests/install_quality/` is empty. Defaults derive from `Path(__file__).resolve().parents[3]`, env-overridable via `REAL_FIXTURE_BASE`; example sidecar uses a `/absolute/path/to/...` placeholder and is still valid JSON.
- **Item 3 — merge, not replace.** `merge()` overrides only keys present in the patch; `children` replaced only when the patch carries a `children` key, so a pin-only override preserves existing `children`/`source`/`model`. A full `RealFixtureSpec` value still whole-spec-replaces (back-compat — `children` resets to `()`). Mutation-probed: reverting the existing-bucket branch to `from_dict` fails `test_override_pin_merges_preserving_children`.
- **Item 5 — nested child paths.** `clone_at` does `dest.parent.mkdir(parents=True, exist_ok=True)`; a `packages/api` child clones instead of skipping, and an uncreatable parent (path element is a file) raises a clean `MissingFixtureError` with `"parent directory"` in the message rather than a cryptic git error.

Rename `test_real_provisioner.py → test_real_provision.py` is clean (no orphan). Item 4 correctly deferred to #101 (warning retained).

Tests: `test_real_provision.py` 24/24; full `framework/tests/` **804 passed**. Mutation-robust on the finally-guarantee and the merge branch.

Must-fix: None
Tech-debt: A brand-new-bucket partial patch that omits `source` raises a bare `KeyError: 'source'` from `from_dict` rather than a friendly `MissingFixtureError`. Pre-existing (the sidecar path already called `from_dict` unconditionally before this PR) and documented in the docstring — noting only so a future pass can turn it into a clear operator-facing error.

PR #248, opened 2026-07-08
harness: provisioner error-hardening + doc note (#243/#244/#242)

parametrization:
## Summary
- **#243** — a `--real-config` / `real_fixtures` partial patch that CREATES a new bucket (one not in the defaults) with no `source` now raises a friendly `MissingFixtureError` naming the bucket, instead of the bare `KeyError('source')` the unconditional `from_dict` raised (runner degrades it to a skip).
- **#244** — strengthen the degenerate zero-children `meta-and-children` guard. Default stays warn-and-proceed (the legitimate bare `--include-real` B10 smoke-test flow should degrade, not crash); a real run opts into a HARD guard via `opts['real_require_children']`, which raises `MissingFixtureError` (runner skip) rather than silently provisioning an empty meta install. Branch extracted into `_guard_zero_children`.
- **#242** — one-line note in `framework/config/README.md`: the amend/upgrade path's `reconcile_module_lists()` deliberately resets the framework-owned hook lists (`pre_bash`/`post_bash`/`post_file`/`session_start`/`agent`/`stop`) to canonical on re-install (user entries dropped); `hooks.pre_push_commands` is the escape hatch.

**#244 design decision (justification):** kept-and-strengthened rather than a blanket hard error. A bare `--include-real` B10 with no `--real-config` is a legitimate smoke-test flow — a childless meta is a valid-if-trivial install — so hard-erroring it by default would break that flow. The stronger guard is therefore opt-in: real provisioning runs that MUST have children (#101/#109) set `real_require_children` and get a visible `MissingFixtureError`; everyone else keeps the clearly-surfaced warning.

**Load-bearing tests** (both fail on revert, verified):
- `test_new_bucket_partial_patch_without_source_friendly_error` — revert the `'source' not in override` guard → bare `KeyError` (not `MissingFixtureError`) → fails.
- `test_meta_zero_children_strict_guard_raises` (+ `_default_off` control) — revert the `real_require_children` branch → warn-only, no raise → `pytest.raises` finds nothing → fails.

Docs-only file (`config/README.md`) is single-homed (framework source of truth, not a byte-mirrored tree); `reinstall.py --check` is green — no source↔runtime drift. Full suite: 812 passed; `ruff check framework/` clean.

## Related Issues
Closes #243
Closes #244
Closes #242

## Review Checklist
- [ ] Must-fix items resolved

### DIFF
--- a/framework/config/README.md
+++ b/framework/config/README.md
@@ -36,6 +36,13 @@ capability but do not error.
 | `hooks.stop` | Stop-event checks (advisory, never block). Ships with `session_handoff` (auto handoff note; defers to a manual `/handoff`). |
 | `hooks.pre_push_commands` | Checks the enforce-mode pre-push git hook runs (read at push time; empty = pass). |
 
+> **Re-install RESETS the framework-owned hook lists to canonical.** Re-installing over an
+> existing config (the amend/upgrade path) reconciles the framework-owned hook module-lists
+> (`hooks.pre_bash`/`post_bash`/`post_file`/`session_start`/`agent`/`stop`) back to the shipped
+> canonical set — stale/extra entries you added there are **dropped**, not merged. Put your own
+> push-time checks in `hooks.pre_push_commands`, which is deliberately excluded from this reconcile
+> and preserved verbatim — it is the user escape hatch for custom checks.
+
 ## How reads work
 
 Hooks call `_framework_config.config(input_data).get("dotted.key", default)`.
--- a/framework/harness/real_provision.py
+++ b/framework/harness/real_provision.py
@@ -259,12 +259,38 @@ def _write_meta_yaml(wd: Path, spec: RealFixtureSpec) -> Path:
     return yaml
 
 
+def _guard_zero_children(spec: RealFixtureSpec, opts: dict) -> None:
+    """#155 item 4 / #244: handle a ``meta-and-children`` spec provisioned with no children.
+
+    A bare ``--include-real`` B10 with no ``--real-config`` resolves to a degenerate
+    zero-children meta install. By DEFAULT this is surfaced with a ``warnings.warn`` and the
+    trivial install proceeds — a childless meta is a valid (if trivial) install, and a smoke-test
+    ``--include-real`` run should degrade rather than crash. A real provisioning run that MUST have
+    children (``#101``/``#109``) can opt into a HARD guard with ``opts['real_require_children']``:
+    the harness then raises ``MissingFixtureError`` (degraded to a *skip* by the runner) instead of
+    shipping the degenerate zero-children install, so a misconfigured real run fails visibly rather
+    than silently provisioning an empty meta.
+    """
+    if opts.get("real_require_children"):
+        raise MissingFixtureError(
+            f"real fixture {spec.bucket!r} is meta-and-children with zero children and "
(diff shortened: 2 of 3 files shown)


parametrization:
Requestor: Ibrahim.El-Amin
Requestee: Paloma.Gupta
RequestOrReplied: Replied

**Review: Clean — friendly source-less error, opt-in zero-children guard, and doc note all correct and load-bearing; ships as-is.**
Must-fix: None
Tech-debt:
1. `opts['real_require_children']` (the #244 hard guard) has no operator-facing surface in `framework/harness/__main__.py` — `--include-real`/`--real-config` populate `include_real`/`real_config` only, so today the strict guard is reachable only by a programmatic caller. The future real runs (#101/#109) will need it wired to a CLI flag or a `--real-config` key. Non-blocking (mechanism + friendly errors are the PR's job and are done + tested); noting for the #101/#109 wiring.

Verification done in an isolated worktree (detached at PR head 92746e5):
- `pytest framework/tests/test_real_provision.py` → 27 passed; `ruff check` clean.
- Revert-to-red proofs: dropped the `'source' not in override` branch → `test_new_bucket_partial_patch_without_source_friendly_error` fails with bare `KeyError('source')`; dropped the `real_require_children` branch in `_guard_zero_children` → `test_meta_zero_children_strict_guard_raises` fails ("DID NOT RAISE"). Both paired tests are genuinely load-bearing. (`_strict_guard_default_off` is a control for unchanged warn behavior — correctly stays green.)
- Correctness: #243 guard is scoped, not over-broad — whole-spec replacements take the `isinstance` branch, existing buckets `merge` (source preserved), only source-less NEW buckets raise. #244 stays default warn-and-proceed for the bare childless smoke-test flow and raises only when opted in.
- Single-homed docs claim verified: `config/README.md` has no `.claude/` mirror and `reinstall.py --check` exits 0. Identity trailers (Paloma + Claude) present on both commits; no bypass.

parametrization:
Requestor: Tariq.Morales
Requestee: Paloma.Gupta
RequestOrReplied: Replied

**Review: provisioner error-hardening (#243/#244) + doc note (#242) — clean, both fixes load-bearing on revert**
Must-fix: None
Tech-debt: None

Verified in a detached worktree at the PR tip:
- Ran `pytest framework/tests/test_real_provision.py -k "zero_children_strict_guard or new_bucket_partial_patch"` — 3 passed.
- Proved both fixes load-bearing by reverting the production hunks: neutralizing the `real_require_children` raise in `_guard_zero_children` (#244) and dropping the `'source' not in override` guard in `_apply` (#243) both flip their tests red (`test_meta_zero_children_strict_guard_raises` no longer raises; `test_new_bucket_partial_patch_without_source_friendly_error` gets a bare `KeyError('source')` instead of `MissingFixtureError`).
- #244 default-off reasoning is sound: warn-and-proceed by default (bare `--include-real` smoke-test still works), hard guard is strictly opt-in via `opts['real_require_children']`; `test_..._default_off` is the control.
- #242 doc note is accurate against code: the six keys listed match `_RECONCILED_HOOK_LISTS = (pre_bash, post_bash, post_file, session_start, agent, stop)` in `bootstrap.py`, and `pre_push_commands` is correctly excluded/preserved (issue prose said "five" — the PR got the real count right).

PR #249, opened 2026-07-08
fix(node): dedupe on bare names + seedable RNG, drop CI retry (#234)

parametrization:
## Summary
- **Dedupe root cause (the real bug):** `addMember`/`randomizeMember` built `generateName`'s `used` set from ROLE-PREFIXED roster *filenames* (`senior_engineer_jane_doe.md` → `"senior engineer jane doe"`), so `used.has("First Last")` never matched a real member and a generated duplicate was silently accepted. New `usedNamesFromRoster` parses the bare `**Name:**` field out of each roster card; both callers now route through it.
- **Seedable RNG:** name/persona pickers called `Math.random()` directly → nondeterministic runs. Introduced a single injectable generator (`setRng` / `resetRng` / `makeSeededRng` — mulberry32); production still defaults to `Math.random`, tests seed for reproducibility.
- **Quarantine removed:** dropped the #230 `--retry=2` vitest quarantine from `.github/workflows/ci.yml` now that the root cause is fixed.

## Determinism evidence
- Suite run **15/15 clean** locally (`npx vitest --run`), zero flakes, before removing the retry.
- Revert-check: reverting the dedupe fix makes both load-bearing dedupe tests FAIL (2 failed / 135 passed); restored → 137 passed.

## Load-bearing tests (fail if the fix is reverted)
- `usedNamesFromRoster` parses bare `First Last`, not the role-prefixed filename.
- `generateName` rejects a candidate already on the roster (scripted RNG forces the collision).
- `addMember(random)` does not recreate a name already on the roster (2 cards post-fix vs 1 pre-fix).
- Seeded RNG: same seed yields the same `generateName` sequence; a full seeded team generation is identical run-to-run and collision-free.

## Related Issues
Closes #234

## Review Checklist
- [ ] Must-fix items resolved

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,18 +80,8 @@ jobs:
 
       - name: Test
         working-directory: node
-        # QUARANTINE (#230): the node suite is nondeterministic. `generateName`
-        # in src/bootstrap.ts dedupes candidate names against a `used` set built
-        # from ROLE-PREFIXED roster filenames ("senior_engineer_jane_doe" ->
-        # "senior engineer jane doe"), never against bare "First Last" names, so
-        # it never actually rejects a real duplicate. Unseeded Math.random() name
-        # generation therefore collides at random, surfacing as a spurious red on
-        # whichever matrix shard happens to draw the collision (W13: "node (20)";
-        # the shard is incidental, not node-20-specific). `--retry=2` re-runs a
-        # failed test up to twice: a genuine, reproducible break still fails all
-        # attempts and goes red (node coverage is preserved and still required),
-        # while a one-off collision self-heals instead of manufacturing a false
-        # ci-red that slips a merge. The real fix — seed the RNG / dedupe on bare
-        # names, then REMOVE this --retry — is tracked as #234 (product change in
-        # node/src, out of this QA story's lane).
-        run: npm test -- --run --retry=2
+        # The #230 --retry=2 quarantine was removed in #234 once the root cause
+        # was fixed: generateName now dedupes candidates against bare "First Last"
+        # names parsed from each roster card (not role-prefixed filenames), and
+        # the RNG is seedable, so the suite is deterministic and needs no retry.
+        run: npm test -- --run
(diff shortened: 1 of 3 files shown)


parametrization:
Requestor: Paloma.Gupta
Requestee: Ibrahim.El-Amin
RequestOrReplied: Replied

**Review: Root-cause fix is correct and load-bearing; determinism verified — ships**
Must-fix: None
Tech-debt:
1. `usedNamesFromRoster` falls back to the legacy filename-derived string when a card's `**Name:**` field can't be parsed — for such a card the dedupe silently reverts to the ineffective form. Deliberate and documented ("better than dropping it"), but worth a follow-up to at least warn on an unparseable roster card so the gap is visible rather than silent. Non-blocking.

parametrization:
Requestor: Tariq.Morales
Requestee: Ibrahim.El-Amin
RequestOrReplied: Replied

**Review: node flake root-fix (#234) — dedupe now on bare names, RNG seedable/prod-default Math.random; determinism confirmed, all four fixes load-bearing**
Must-fix: None
Tech-debt:
1. `usedNamesFromRoster` falls back to the legacy role-prefixed filename string when a card's `**Name:**` can't be parsed — that fallback re-introduces the exact #234 mismatch for an unparseable card (a role-prefixed string never matches a bare `First Last` candidate). Harmless for all real roster cards (which carry `**Name:**`), so it does not defeat the shipped guarantee — noting as non-blocking follow-up only.

Verified in a detached worktree at the PR tip:
- Ran `npx vitest --run` — 137 passed. Ran it 5x back-to-back: 137/137 every time (determinism, the point of #234).
- Dedupe genuinely compares bare names now: reverting `usedNamesFromRoster` to the old filename-derived set flips `usedNamesFromRoster parses bare First Last` and the end-to-end `addMember(random) does not recreate a name already on the roster` red.
- RNG injection load-bearing + prod default correct: `_rng` defaults to `Math.random` (no seeding at module load — production unchanged); reverting `randomChoice` to call `Math.random()` directly flips both `seedable RNG is reproducible` tests red.
- CI change is correct: `--retry=2` quarantine (#230) removed, node coverage still required (`npm test -- --run`).

PR #268, opened 2026-07-08
W18 S3: Provisioner + node warn-surface — real_require_children CLI (#251) + usedNamesFromRoster unparseable-card warn (#252)

parametrization:
## W18 S3 — Provisioner + node warn-surface (two small, disjoint hardening items)

Closes #251
Closes #252

### #251 — `real_require_children` exposed via `--real-config` schema
The zero-children hard guard was programmatic-only (`opts['real_require_children']`,
shipped W11 #244). This adds a per-bucket `require_children: bool` field to
`RealFixtureSpec` (wired through `to_dict`/`from_dict`/`merge`) so an operator can
toggle the guard from the `--real-config` sidecar JSON without editing code, e.g.:

```json
{ "B10": { "source": "…", "model": "meta-and-children", "require_children": true } }
```

`_guard_zero_children` now fires on `opts.get("real_require_children") or
spec.require_children`. **Default-off opt-in is preserved** (#244): absent key →
`False` → warn-and-proceed; the hard `MissingFixtureError` only fires when explicitly
enabled by either surface. (The provisioner CLI's `__main__.py` is another story's
file and untouched — the `--real-config` schema is the "and/or" surface chosen here.)

### #252 — `usedNamesFromRoster` unparseable-card warn (node)
When a roster card's `**Name:**` field can't be parsed, `usedNamesFromRoster` now
emits a visible `console.error` warning naming the card + the fallback string,
**before** taking the legacy filename-derived fallback. The fallback itself is
unchanged (the #234 dedupe / seedable-RNG path is untouched) — the degrade is just
no longer silent.

### Tests + revert→red evidence
Python (`framework/tests/test_real_provision.py`):
- `test_require_children_surfaced_via_real_config_schema` — sidecar `require_children`
  drives the raise with **no** global opts flag; embedded in-memory patch strips the
  `spec.require_children` disjunct and confirms it degrades to a warn (no raise).
  **Verified revert→red:** reverting the production `or spec.require_children` in
  memory drops the raise → the part-(a) `pytest.raises` would fail.
- `test_require_children_defaults_off_and_round_trips` — default-off, `merge`
  preserve/override, `to_dict`/`from_dict` round-trip.

Node (`node/tests/cli.test.ts`):
- `warns on stderr when a card's **Name:** cannot be parsed (#252)` — asserts the
  fallback still lands in the set AND a warning is emitted; embeds an in-memory
  reimplementation without the warn to demonstrate the assertion. **Verified
  revert→red:** temporarily removing the production `console.error` makes
  `expect(errSpy).toHaveBeenCalledTimes(1)` fail (1 failed), restored → green.
- `does not warn when **Name:** parses cleanly (#252 control)`.

### Status
- Python: `ruff check` clean; `pytest framework/tests/test_real_provision.py -q` →
  29 passed.
- Node: `npm run build` clean; `npm test` → 139 passed (5 files).
- Files changed (4, exactly the allowed set): `framework/harness/real_provision.py`,
  `framework/tests/test_real_provision.py`, `node/src/bootstrap.ts`,
  `node/tests/cli.test.ts`. `node/src` is not a `.claude/` asset — no dual-tree parity.

### DIFF
--- a/framework/harness/real_provision.py
+++ b/framework/harness/real_provision.py
@@ -86,11 +86,13 @@ class RealFixtureSpec:
     ref: str = "refs/heads/main"
     model: str = "single-repo"                 # "single-repo" | "meta-and-children"
     children: tuple[RealChildSpec, ...] = ()
+    require_children: bool = False             # #251: per-bucket opt-in zero-children hard guard
 
     def to_dict(self) -> dict:
         return {"bucket": self.bucket, "source": self.source, "pin": self.pin,
                 "ref": self.ref, "model": self.model,
-                "children": [c.to_dict() for c in self.children]}
+                "children": [c.to_dict() for c in self.children],
+                "require_children": self.require_children}
 
     @classmethod
     def from_dict(cls, bucket: str, d: dict) -> RealFixtureSpec:
@@ -98,6 +100,7 @@ class RealFixtureSpec:
             bucket=bucket, source=d["source"], pin=d.get("pin"),
             ref=d.get("ref", "refs/heads/main"), model=d.get("model", "single-repo"),
             children=tuple(RealChildSpec.from_dict(c) for c in d.get("children", ())),
+            require_children=bool(d.get("require_children", False)),
         )
 
     def merge(self, d: dict) -> RealFixtureSpec:
@@ -111,6 +114,9 @@ class RealFixtureSpec:
         children = self.children
         if "children" in d:
             children = tuple(RealChildSpec.from_dict(c) for c in d["children"])
+        require_children = self.require_children
+        if "require_children" in d:
+            require_children = bool(d["require_children"])
         return replace(
             self,
             source=d.get("source", self.source),
@@ -118,6 +124,7 @@ class RealFixtureSpec:
             ref=d.get("ref", self.ref),
             model=d.get("model", self.model),
             children=children,
+            require_children=require_children,
         )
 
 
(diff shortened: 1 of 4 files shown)


parametrization:
Requestor: Paloma.Gupta
Requestee: Ibrahim.El-Amin
RequestOrReplied: Replied
Must-fix: None
Verified:
- Python static+suite: `ruff check framework/harness/real_provision.py` clean; `pytest framework/tests/test_real_provision.py -q` -> 29 passed.
- Node build+suite: `npm run build` clean (tsc, no errors); `npm test` -> 139 passed (5 files).
- Revert->red #251 (in-memory patch, no git checkout): stripped `or spec.require_children` from production `_guard_zero_children` -> `test_require_children_surfaced_via_real_config_schema` went RED ("DID NOT RAISE MissingFixtureError" at line 376); restored -> green. Confirms the disjunct is what drives the guard.
- Revert->red #252 (in-memory patch): stripped the `console.error` in production `usedNamesFromRoster`, rebuilt -> `warns on stderr ... (#252)` went RED ("expected error to be called 1 times, got 0"); #234 dedupe/RNG tests + the #252 clean-parse control stayed green; restored -> green.
- #251 end-to-end sidecar toggle (independent probe, real git meta bucket, opts={}): sidecar `{"B10":{"model":"meta-and-children","require_children":true}}` on a childless meta -> `real_registry` parses `spec.require_children=True` and `provision_real` RAISES MissingFixtureError with NO global `real_require_children` opt. Default (absent key) -> `require_children=False`, same childless meta only WARNS and proceeds (#244 default-off genuinely preserved). Operator-toggleable without code edits -> #251 AC ("CLI and/or --real-config surface") genuinely satisfied.
- #251 schema plumbing: `to_dict` emits the key, `from_dict(to_dict())` round-trips True, `merge({"pin":...})` preserves an existing True, `merge({"require_children":false})` overrides to False -> no drop through any path.
- #252 routing + non-regression (independent probe, stdout/stderr split): warning goes to STDERR only (not stdout), names the card file + the filename-derived fallback; the fallback name still lands in the dedupe set (behavior unchanged) and a clean `**Name:**` card still dedupes on the bare name (#234 logic untouched). No CLI flag added to `__main__.py` (correct file-disjointness).

parametrization:
Requestor: Nia.Rossi
Requestee: Ibrahim.El-Amin
RequestOrReplied: Replied
Must-fix: None
Verified:
- Python: `ruff check framework/harness/real_provision.py` clean; `pytest framework/tests/test_real_provision.py -q` = 29 passed.
- Node: `npm run build` clean (tsc); `npm test` = 139 passed (cli 87 / personas 14 / copy-shared 5).
- Revert->red #251: in-memory dropped `or spec.require_children` from `_guard_zero_children` -> `test_require_children_surfaced_via_real_config_schema` RED (DID NOT RAISE MissingFixtureError); restored -> green.
- Revert->red #252: in-memory stripped the `console.error` in `usedNamesFromRoster` -> `warns on stderr when a card's **Name:** cannot be parsed (#252)` RED (toHaveBeenCalledTimes(1) failed); restored+rebuilt -> green.
- Config-plumbing (TL lens): traced `require_children` from_dict->merge->to_dict directly. Default-off (#244) preserved: no key -> False, warn-and-proceed (no raise on zero children). Round-trip `from_dict(to_dict(x))` retains True. `merge` PRESERVES on a keyless partial patch (True->True, False->False) and OVERRIDES both directions when the key is present (True->False, False->True) — sidecar surface is reliable, no silent drop-to-default.
- #252 surface: warning is `console.error` (stderr, bootstrap.ts:208), not `console.log`/stdout; filename-derived fallback + #234 bare-name dedupe/RNG paths unchanged; control test "does not warn when **Name:** parses cleanly" green.

PR #276, opened 2026-07-08
fix(hooks): guard non-dict rulesets `parameters` in CI-status probe (#269)

parametrization:
## S3 of Wave 19 (Phase 6 W14) — guard non-dict rulesets `parameters` (#269)

Fixes tech-debt **#269** in `_rulesets_enforce_required_checks`
(`framework/assets/hooks/validate_pr_ci_status.py`, the rulesets CI-status probe I
wrote in W18).

### The bug
The scan read `params = rule.get("parameters") or {}`. The GitHub rulesets API can
return a rule's `parameters` as a non-empty **list** rather than a dict. A truthy
list slips past `or {}`, and the next `params.get(...)` raises `AttributeError` —
which crashes the whole scan into fail-open. Safe (no false block), but noisy, and
it **discards the real enforcement info** carried by the other well-formed rules in
the same payload.

### The guard
```python
params = rule.get("parameters")
if not isinstance(params, dict):
    continue
```
A rule whose `parameters` is not a mapping carries **no enforcement info**: it is
skipped, never raised on. The scan continues to the remaining rules, so genuine
enforcement configured by a well-formed sibling rule is still recognized.

### Tri-state preserved exactly
The fail-open tri-state is unchanged in every path:
- **True** — happy path, classic-enforces, rulesets-enforces (incl. a good rule
  after a malformed one)
- **False** — empty rules / 404 / empty required set (definitive not-enforced); a
  malformed rule never manufactures a false "not enforced"
- **None** — no repo/base, transport/permission error, bad JSON, non-list payload
  (the fail-open sentinel meaning "could not determine")

A malformed rule is treated as "no info from this rule," never as "not enforced."

### Tests (added to the existing hook suite; both revert→red against the old form)
- `test_rulesets_list_parameters_does_not_raise_or_false_enforce` — list
  `parameters` does not raise and yields a **definitive False** (no other enforcing
  rule). Reverting the guard reproduces the exact `AttributeError` of #269.
- `test_rulesets_list_parameters_preserves_other_rule_enforcement` — a list-params
  rule is skipped but a well-formed enforcing rule after it is still honored → True.

### Verification
- Full hook suite: **42 passed** (all prior fail-open tests green).
- `ruff check` clean (CI runs only `ruff check`).
- `python3 framework/install/reinstall.py --check` → **rc=0** (single-source hook,
  wired in place via `.claude/settings.json`; no `.claude/hooks/` mirror — no drift).

Closes #274
Closes #269

### DIFF
--- a/framework/assets/hooks/validate_pr_ci_status.py
+++ b/framework/assets/hooks/validate_pr_ci_status.py
@@ -342,7 +342,15 @@ def _rulesets_enforce_required_checks(repo: str | None, base: str | None) -> boo
             continue
         if rule.get("type") != "required_status_checks":
             continue
-        params = rule.get("parameters") or {}
+        params = rule.get("parameters")
+        # The rulesets API can return `parameters` as a non-dict (e.g. a list),
+        # on which `.get(...)` would raise AttributeError and crash into
+        # fail-open — safe but noisy, and it discards the real enforcement info
+        # carried by the OTHER well-formed rules. A rule whose parameters we
+        # cannot read as a mapping simply carries NO enforcement info: skip it,
+        # don't raise, and don't let it manufacture a false "not enforced".
+        if not isinstance(params, dict):
+            continue
         required = params.get("required_status_checks") or []
         if required:
             return True
--- a/framework/tests/test_validate_pr_ci_status.py
+++ b/framework/tests/test_validate_pr_ci_status.py
@@ -196,6 +196,21 @@ _RULESET_EMPTY_CHECKS = (
     '[{"type": "required_status_checks", "ruleset_id": 7, '
     '"parameters": {"required_status_checks": []}}]'
 )
+# A required-status-checks rule whose `parameters` is a non-dict (a LIST). The
+# rulesets API can shape it this way; `.get(...)` on a list would raise (#269).
+_RULESET_LIST_PARAMS = (
+    '[{"type": "required_status_checks", "ruleset_id": 7, '
+    '"parameters": [{"required_status_checks": [{"context": "node (20)"}]}]}]'
+)
+# A LIST-parameters rule (must be skipped, not raise) sitting BEFORE a
+# well-formed enforcing rule — proves the guard preserves the other rule's
+# enforcement info instead of crashing the whole scan into fail-open.
+_RULESET_LIST_PARAMS_THEN_ENFORCING = (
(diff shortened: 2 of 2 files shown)


parametrization:
Requestor: Ibrahim El-Amin
Requestee: Nia Rossi
RequestOrReplied: Replied
Must-fix: None
Verified:
- revert->red: edited the guard back to `params = rule.get("parameters") or {}` and ran `test_rulesets_list_parameters_does_not_raise_or_false_enforce` in isolation — it went RED at line 346 with the exact #269 `AttributeError: 'list' object has no attribute 'get'`; restored the guard and the test is green again.
- Full suite green: `pytest framework/tests/test_validate_pr_ci_status.py` = 42 passed with the guard in place.
- Non-dict shapes all skip safely (no raise, no manufactured enforcement): direct-drove `_rulesets_enforce_required_checks` with `parameters` = list / null / str / int — every one returns a DEFINITIVE False (not None, not a crash) when no other rule enforces.
- Tri-state + sibling-rule preserved: enforcing dict -> True, empty `required_status_checks` -> False, and a list-params rule sitting BEFORE a well-formed enforcing rule still -> True (the malformed rule is skipped, real enforcement honored).
- ruff clean on both changed files.

parametrization:
Requestor: Tariq Morales
Requestee: Nia Rossi
RequestOrReplied: Replied
Must-fix: None
Verified:
- revert->red: reverted the `isinstance(params, dict)` guard to pre-#269 `rule.get("parameters") or {}` and `test_rulesets_list_parameters_does_not_raise_or_false_enforce` went RED with `#269 AttributeError: 'list' object has no attribute 'get'` at validate_pr_ci_status.py:346; restored guard -> green.
- Full hook suite: 42 passed (matches author's report); zero-diff working tree after restore.
- Guard robustness probe: `parameters` as list/null/str/int/empty-dict all skip cleanly — no raise, no false enforce (each -> definitive False with no other enforcing rule); a well-formed enforcing rule AFTER a list-params rule still returns True (preserves other-rule enforcement).
- Tri-state unchanged: well-formed enforcing -> True; empty required set -> False; API-unavailable/bad-JSON -> None sentinel (fail-open) all intact.
- reinstall.py --check rc=0: hook is single-source (wired in place, no `.claude/hooks/` mirror), canonical `.claude/` in sync.

PR #306, opened 2026-07-09
fix(trust): render every negative signal + reject blank receipts

parametrization:
## Summary

The forced negative-signal pass — the mechanism whose entire job is to ban vacuous
entries — both **emitted** and **accepted** a blank receipt for a `missed_catches`-
or `gate_bypasses`-only wave, so the scorer applied a `-1` it could not cite.

Root cause: **two hand-maintained parallel field lists**. `Signals.has_negative()`
tested six negative fields; `negative_signal_line()`'s `gaps` list rendered only four.
`missed_catches` and `gate_bypasses` fell through to a blank `"Name: "`, and
`validate_negative_signal_pass()` waved that blank through.

- **One source of truth** — `_NEGATIVE_SIGNAL_FIELDS` (field name + gap template
  together). Both `has_negative()` and `negative_signal_line()` derive from it, so the
  field set and its renderer **cannot drift again**: a negative added there — and
  nowhere else — both counts and renders.
- `validate_negative_signal_pass()` now rejects an entry **blank after the `"{name}: "`
  prefix** (and the wholly-blank line), not just bare `None`/`n/a`/`-`.
- `apply_distribution_discipline` docstring now states that **ties at the top composite
  are intended** (per the #275 rationale: `difficulty_points` was author-only and
  dominated the composite, so a flagship author always outranked a defect-catching
  reviewer; weighting review value at `REVIEW_VALUE_WEIGHT` and capping difficulty at
  `DIFFICULTY_COMPOSITE_CAP` fixed it). A future reader must not "fix" the reserved 5
  into a single seat. (Satisfies the criterion from #301 that lives in this file.)

Trackers this addresses: #296 (the closing keyword is in the **commit trailer**, not
this body — deliberately, so this PR cannot re-trigger the #304 auto-close footgun that
already bit #296 via PR #299), plus the docstring criterion from #301.

## Before / after (reproduction from the issue)

```
# before (main @ 8b43481)
negative_signal_line('Ibrahim El-Amin', Signals(missed_catches=1, ...))  ->  'Ibrahim El-Amin: '   # blank
validate_negative_signal_pass(['Ibrahim El-Amin: '])                     ->  []                    # accepted (bug)
negative_signal_line('X', Signals(gate_bypasses=1))                      ->  'X: '                 # blank

# after (this PR)
negative_signal_line('Ibrahim El-Amin', Signals(missed_catches=1, ...))  ->  'Ibrahim El-Amin: 1 missed catch(es) as reviewer'
[...]

PR #305, opened 2026-07-09
docs(wave-retro): pluralize reserved-5 to "top relative performer(s)"

parametrization:
## Summary

- Pluralize the reserved-5 wording to **"top relative performer(s)"** in the three normative sites, reconciling the `/wave-retro` skill and trust-matrix note with `apply_distribution_discipline` (owner decision 2026-07-09). **No behavior change** -- `lib/trust_signals.py` is untouched; the code already seats every engineer at the strictly-positive iteration-max composite, which is the author-vs-reviewer parity #275 was built to produce.
- `/wave-retro` step 5: document that a tie at the top composite is a **reportable finding, not a blocker** -- seat all tied engineers, record their composites; note the rejected fewest-negatives tie-break.
- `/wave-retro` step 7: proposal-writing guidance forbidding a leading closing-keyword + issue reference (satisfies #304's proposal-guidance criterion; wording mirrors the hook regex in #304).

## Related work

Addresses the wording reconciliation tracked in `#301` -- the closing keyword is intentionally kept **out** of this PR body (the `Closes` trailer lives in the commit message) precisely because a quoting PR body auto-closes issues on merge, the defect #304 exists to prevent. Also lands #304's `/wave-retro` proposal-guidance criterion (I own every `skills/wave-retro/` edit this wave).

## Normative vs historical (the AC correction)

Per the pinned AC correction on the issue, only **normative** text is pluralized:

- Changed: `.claude/skills/wave-retro/SKILL.md:176`, its `framework/assets/` mirror, and one blockquote note in `.claude/team/trust_matrix.md` (was line 658).
- Left **byte-for-byte untouched** (historical -- they record the rule as it was when those scores were assigned): `trust_matrix.md:197` (Wave-13 reason cell) and every `feedback_log.md` hit. A grep or test forcing these to change is itself the defect. `git diff` shows zero change to line 197.

### Reviewer call-out: this is a BACK-ANNOTATION of a Wave-22 retro note, not a preamble edit

Be precise about what the `trust_matrix.md` edit touches, because this wave's theme is "the record must not lie." The issue and my commit message call the changed line the "preamble note," but structurally that blockquote sits **inside the Wave 22 Trust Updates section** (immediately below the `Signals via trust_signals.py extract/score 22` line), not in the matrix's top-of-file preamble. So I am editing a **historical retro note** -- the owner explicitly designated it normative in the pinned AC correction, but reviewers should consciously sign off on touching a Wave-22 record.

[...]

PR #292, opened 2026-07-08
fix(trust): make trust_signals + review_load author-exclusive (#288)

parametrization:
## #288 — make the trust scorer + review-load author-exclusive

The merge gate ignores a PR author's own verdict on their own PR, but
`trust_signals.py` and `review_load.py` did not. In Wave 21 an author's must-fix
reply wore reviewer verdict grammar (`Requestor: <self>` on their own PR, #287),
so the scorer counted the author as a third reviewer of their own PR →
spurious `missed_catches` / `verified_reviews` (worked around by amending). This
fixes it durably.

### Shared helper
`trust_signals.is_author_self_review(requestor, author, canon) -> bool` — true
when a verdict's `Requestor:` canonicalizes to the PR's own author. The
charter's `Requestor:` names the *reviewer*, and a reviewer reviews someone
else's PR, so such a verdict is the author wearing reviewer grammar on their own
PR — never a genuine review. Fail-open/total (missing `requestor` or `author` ⇒
not a self-review); both identities fold through `canon` so name variants
(`Ibrahim.El-Amin` / `Ibrahim El-Amin (Senior)`) compare equal. `review_load`
already imports `trust_signals as ts`, so both surfaces share this one helper
(no new module).

### Surfaces
- **`_account_pr`**: author self-verdicts are dropped from the parsed `verdicts`
  and the `current_verdicts` (gate-bypass) sets before any accounting, and a
  durable ledger catch whose `Requestor:` is the author is skipped. No reviewer
  signal (`must_fix_caught` / `verified_reviews` / `missed_catches` /
  false-positives / the #H5 clean-reviewer count) is ever attributed to the
  author for their own PR.
- **`review_load`**: the pure core now takes `(pr_key, author, comment_bodies)`
  and author-excludes via the same helper; `extract_review_load` threads
  `commit_author_name`. Unknown (`None`) author excludes nothing (fail-open).

### Charter
`team/charter/issues.md` verdict-grammar section now states the grammar is
**reviewer-only** — a PR author's reply to a must-fix is a PLAIN comment, never
a `Requestor:` / `RequestOrReplied:` verdict — with the scorer's author-exclusion
called out as the durable backstop. Dual-tree: canonical template + live
re-rendered through `install_charter` (manifest refreshed).

### Scope — SCORING surfaces only (this fix does NOT touch the merge gate)
#288 makes the two **scoring** surfaces (`trust_signals` + `review_load`)
author-exclusive. It does **not** change the merge gate. An earlier draft of this
[...]

PR #290, opened 2026-07-08
docs(intake): 2026-07 sibling-repo mining findings (#264)

parametrization:
## Story S1 / Wave 22 (Phase 7) — #264 sibling-repo mining audit

Docs-only. Re-audits `botfarm_inc` + `noorinalabs-main` `.claude/`/`ontology/` changes since the
2026-06-24 mine and folds the ranked candidates into the #102 backlog. **No product code, hooks,
or assets modified** — audit + triage only.

Adds `intake/2026-07/MINING-FINDINGS.md`: summary + classification tally, per-repo candidate
tables (classification / net-new-vs-refinement / P-tier / rationale), a ranked port queue, and the
required **#657 ↔ #288 cross-check**.

### Candidate count (~38)
| Classification | Count |
|---|---:|
| port-as-is | 2 |
| genericise-then-port | 15 |
| already-covered — refinement | 13 |
| already-covered — verify-first | 4 |
| skip / guidance-only | 4 |

### Top P-tier items
- **P0** noorinalabs #881 `trust_signals` false-positive scoping (refinement, drop-in)
- **P1** botfarm `_resolves_to_author` helper shape (land WITH #288)
- **P1** noorinalabs #864 `verify_deployable_merge` (net-new gap)
- **P1** noorinalabs #886/#896/#897 `validate_pr_review` batch-loop hardening
- **P1** botfarm #424 reviewer-load cap as a hard kickoff gate

### #657 ↔ #288 cross-check
botfarm #657 is a review-gate advisory-message refinement (counting unchanged) — **not** the
#288 counting fix. But its `_resolves_to_author` three-anchor helper is exactly #288 prong-1's
"single shared author-resolution helper"; port the shape with a mirrored trigger (2real excludes
when the *Requestor* resolves-to-author). Full reasoning in the note.

### Caveat
2real's `_shell_parse`/`_repo_flag_parse` are more mature than the source snapshots, so several
loud source "gate-evasion" fixes are likely already-covered here (marked verify-first). Diff each
refinement against `framework/assets/` before scoping a port story.

Dogfoods the Wave-21 default-on `docs` load-bearing-test exception: this pure-doc diff passed
`require_load_bearing_test` without tripping the gate.

Backlog comment: parametrization/2real-team-framework#102 (issuecomment-4920210060).
Closes #264.

PR #307, opened 2026-07-09
docs(wave-end): recompute wave counters from trust_signals extract (#300)

parametrization:
## Summary
- `/wave-end` step 3 now **derives** the three wrapup counters mechanically instead of accepting numbers transcribed from the session handoff's prose — the handoff narrates *final verdict state* (an in-place-amended PR reads "clean") while the counters track *round history*, so transcription is a second source of truth that guarantees drift against `/wave-retro` step 2 (bit twice: Wave 22 `--cr-cycles 1` for a two-rework-round wave; Wave 11 → the `origin/<wave>` runbook step #255).
- `--cr-cycles` (`sum(rework_cycles)`) **and** `--concentration` (`max/total` over `authored_prs`) both derive from `trust_signals.py extract {W}` — the identity-aware source the retro reconciles against (it resolves team identity from `Co-Authored-By` trailers). Grouping `gh`'s `.author.login` collapses this dogfooded repo's single account to ~100% every wave, so both identity-sensitive counters avoid it. `--pr-count` stays on `gh pr list | jq length` — a count of PRs, not authors, so it is identity-agnostic and correct as-is.
- Writes the recomputed-vs-claimed reconciliation in **both** directions: recomputed `<` claimed explained by edited-in-place verdicts → claimed stands (corrections entry); recomputed `>` claimed → recomputed wins (the Wave 22 mirror image the old docs never covered).
- **Silent-failure guardrail (three orthogonal defenses).** `extract` fails silent under a rate limit (`{}`, exit 0), and `gh pr list` is GraphQL-backed — the *same* bucket — so under a pre-exhausted budget both fail in lockstep, `${PR_COUNT:-0}` collapses to 0, and a cross-check that also anchors on 0 passes (the #307 review defect). Step 3 now, before deriving anything: **(A)** preflights the GraphQL budget over `gh api rate_limit` (served over REST — a bucket that can't fail in lockstep), aborting near zero; **(B)** captures `gh pr list`'s exit status explicitly (non-zero exit / non-numeric stdout → abort, never a `${VAR:-0}` collapse); **(C)** captures `extract` once and cross-checks its identity-aware `authored_prs` total against the now-trusted `PR_COUNT` (empty or partial → abort). A genuinely empty wave proceeds only after (A)+(B) prove the reads succeeded. Live-verified: at `graphql.remaining=0`, defense (A) aborts and `wave wrapup` is not called.
- Dual-tree (`.claude/` + `framework/assets/`), byte-parity verified; `reinstall.py --check` reports in-sync.

## Follow-up (separate surface — not touched here)
`trust_signals.py extract` reporting an API failure as an empty success (`{}`, exit 0) is the library-level half of this defect, tracked as **#311** on the `trust_signals.py` surface which this story must not touch. The `/wave-end` guard defends the record independently — it never trusts an `extract` result it can't corroborate against an orthogonal read.

## Related Issues
Refs #300 (closing keyword intentionally kept in the commit message, not here — PR-body closers auto-close on merge regardless of outcome, #304).

## Verification
- **Wave-22 replay through the documented procedure yields `cr_cycles = 2` and `concentration = 33`** (`trust_signals.py extract 22`: Nia `rework_cycles=1`/`authored_prs=[290]`, Paloma `rework_cycles=1`/`[292]`, Ibrahim `[294]` → cr-cycles sum 2, concentration `max 1 * 100 // total 3` = 33, matching `wave_22_top_concentration_pct`) — output pasted in the PR thread.
- Regression `framework/tests/test_wave_end_counter_derivation.py`: Wave-22 arithmetic (2 and 33); the abort gate distinguishing the three states (empty-wave → proceed; read-failed, incl. the pre-exhausted lockstep case → abort; reads-disagree → abort) with revert→red confirmed; and content guards that both SKILL.md trees use `authored_prs`, never re-introduce `group_by(.author.login)`, carry all three abort defenses, and call `extract` exactly once.
- `ruff check` clean; framework test suite 927 passed.
- Byte-parity between the two skill trees confirmed via `diff` (identical; no mustache placeholders in the edited section); `reinstall.py --check` in sync.

## Review Checklist
- [ ] Must-fix items resolved

### DIFF
--- a/.claude/skills/wave-end/SKILL.md
+++ b/.claude/skills/wave-end/SKILL.md
@@ -58,41 +58,69 @@ and reads the counters recorded here.
    # wave-branch    → BASE=<branch.integration template, the wave's phase + $W>
    # direct-to-main → BASE=<default branch>; add `--label "wave-$W"` to scope the wave
 
-   # --pr-count is identity-AGNOSTIC — it counts merged PRs, not authors — so the raw
-   # `gh` list is correct here even in a single-account (dogfooded) repo where every PR
-   # carries one `author.login` and team identity lives only in `Co-Authored-By`:
-   PR_COUNT="$(gh pr list --state merged --base "$BASE" --json number | jq 'length')"
+   # REFUSE to derive counters from a read you cannot trust. `extract` fails SILENT
+   # under a GraphQL rate limit (returns `{}`, exit 0), and `gh pr list --json` is
+   # GraphQL-backed too — the SAME 5,000/hr bucket — so when the budget is pre-exhausted
+   # BOTH come back empty and a naive `${PR_COUNT:-0}` collapses to 0, then the
+   # cross-check compares 0 == 0 and reports health (#307 review). Three ORTHOGONAL
+   # defenses run before any counter is derived or `wave wrapup` is called:
+
+   # (A) Preflight the GraphQL budget over an INDEPENDENT channel. `gh api rate_limit`
+   #     is served over REST — a different bucket that does NOT drain with GraphQL — so
+   #     it stays readable when `gh pr list` / `extract` cannot. If GraphQL is at/near
+   #     zero, the wave read WILL return empty-but-successful; abort now, naming the
+   #     limit. `GQL_FLOOR` is a fail-fast margin; (B) and (C) are the hard guarantees.
+   GQL_FLOOR=100
+   GQL_REMAINING="$(gh api rate_limit --jq '.resources.graphql.remaining' 2>/dev/null)"
+   case "$GQL_REMAINING" in
+       ''|*[!0-9]*)
+           echo 'ABORT: could not read the GraphQL rate-limit budget (REST rate_limit failed).' >&2
+           echo '  Cannot establish that the wave read will succeed — refusing to record counters.' >&2
+           exit 1 ;;
+   esac
+   if [ "$GQL_REMAINING" -lt "$GQL_FLOOR" ]; then
+       echo "ABORT: GraphQL budget near-exhausted ($GQL_REMAINING < $GQL_FLOOR)." >&2
+       echo '  gh pr list and trust_signals extract both draw on this bucket and would' >&2
+       echo '  return empty-but-successful, recording FALSE zero counters. Wait for reset —' >&2
+       echo '  gh api rate_limit --jq .resources.graphql — then re-run step 3 from the top.' >&2
+       exit 1
+   fi
+
+   # (B) --pr-count is identity-AGNOSTIC (a count of PRs, not authors, so `gh`'s single
+   #     dogfood `.author.login` is fine here) — but its READ can still fail. Capture the
+   #     exit status EXPLICITLY: a failed `gh pr list` must not collapse into '' and then
+   #     0 via `${VAR:-0}`. A non-zero exit, or non-numeric stdout, is an ABORT, not a
+   #     zero. (This is what the pre-exhausted case trips — `gh pr list` exits non-zero.)
+   PR_LIST="$(gh pr list --state merged --base "$BASE" --json number)"; PR_LIST_RC=$?
+   PR_COUNT="$(printf '%s' "$PR_LIST" | jq 'length' 2>/dev/null)"
+   if [ "$PR_LIST_RC" -ne 0 ] || ! printf '%s' "$PR_COUNT" | grep -Eq '^[0-9]+$'; then
+       echo "ABORT: gh pr list failed (exit $PR_LIST_RC) or returned a non-numeric count." >&2
+       echo '  A failed read is NOT an empty wave — refusing to record zero counters.' >&2
+       exit 1
+   fi
 
    # --cr-cycles and --concentration are BOTH identity-SENSITIVE, so both derive from
-   # `trust_signals.py extract` — the only source that resolves team-member identity
-   # (it parses the `Co-Authored-By` trailers). Grouping `gh`'s `.author.login` instead
-   # collapses every PR into the single bot/owner account in a dogfooded repo, pinning
-   # concentration at ~100% forever — exactly the drift this derivation exists to kill.
-   # Extract once and read both counters off it — a second call could straddle a
-   # rate-limit boundary and return different data:
+   # `trust_signals.py extract` — the only source that resolves team-member identity (it
+   # parses the `Co-Authored-By` trailers); grouping `gh`'s `.author.login` would collapse
+   # a dogfooded repo to ~100% concentration forever (#307). Extract ONCE — a second call
+   # could straddle a rate-limit boundary and return different data:
    SIG="$(python3 "$LIB/trust_signals.py" extract "$W")"
 
-   # GUARDRAIL — `extract` fails SILENT (#300 follow-up; the fix at the `extract`
-   # level is tracked separately on the trust_signals surface). It is SCM-dependent,
-   # and under a GitHub API rate limit it returns `{}` with exit 0 and no stderr. The
-   # `add // 0` / zero-total guards below would then record `cr_cycles=0,
-   # concentration=0` — a PLAUSIBLE-looking lie (a genuinely clean wave IS 0),
-   # indistinguishable from "we could not read GitHub". So the record must not be
-   # written from an unreadable extract. You already know PR_COUNT independently from
-   # `gh pr list`; cross-check it against extract's identity-aware `authored_prs`
-   # total. An empty map — or ANY disagreement, which also catches a PARTIAL read —
-   # for a wave that merged PRs is an ABORT, never a zero:
-   SIG_ENGINEERS="$(echo "$SIG" | jq 'length' 2>/dev/null || echo 0)"
-   SIG_PR_TOTAL="$(echo "$SIG" | jq '[.[].authored_prs | length] | add // 0' 2>/dev/null || echo 0)"
-   if [ "${PR_COUNT:-0}" -gt 0 ] && { [ "${SIG_ENGINEERS:-0}" -eq 0 ] || [ "${SIG_PR_TOTAL:-0}" -ne "${PR_COUNT}" ]; }; then
+   # (C) Cross-check `extract` against the now-TRUSTED PR_COUNT (reads (A)+(B) confirmed
+   #     succeeded). `extract` still fails SILENT (`{}`, exit 0) — the #300 follow-up; the
+   #     library-level fix is #311 on the trust_signals surface, not touched here. An
+   #     empty map — or ANY disagreement, which also catches a PARTIAL read — for a wave
+   #     that merged PRs is an ABORT, never a zero. A genuinely empty wave reaches here
+   #     only after the reads SUCCEEDED, so PR_COUNT == 0 is then a real zero and proceeds:
+   SIG_ENGINEERS="$(printf '%s' "$SIG" | jq 'length' 2>/dev/null || echo 0)"
+   SIG_PR_TOTAL="$(printf '%s' "$SIG" | jq '[.[].authored_prs | length] | add // 0' 2>/dev/null || echo 0)"
+   if [ "$PR_COUNT" -gt 0 ] && { [ "${SIG_ENGINEERS:-0}" -eq 0 ] || [ "${SIG_PR_TOTAL:-0}" -ne "$PR_COUNT" ]; }; then
        echo "ABORT: trust_signals extract could not be trusted for wave $W" >&2
        echo "  gh pr list count = $PR_COUNT, but extract shows $SIG_ENGINEERS engineer(s)" >&2
-       echo "  covering $SIG_PR_TOTAL authored PR(s). An empty/short extract for a wave" >&2
-       echo "  that merged PRs means the READ FAILED (usually a GraphQL rate limit)," >&2
-       echo "  NOT that no rework happened. Do NOT run 'wave wrapup' with these numbers —" >&2
-       echo "  it would write false counters into state.json that nothing downstream" >&2
-       echo "  could tell from real zeros. Wait for reset —" >&2
-       echo "  'gh api rate_limit --jq .resources.graphql' — and re-run step 3 from the top." >&2
+       echo "  covering $SIG_PR_TOTAL authored PR(s). An empty/short extract for a wave that" >&2
+       echo "  merged PRs means the READ FAILED (usually a GraphQL rate limit), NOT that no" >&2
+       echo "  rework happened. Do NOT run 'wave wrapup' with these numbers. Wait for reset" >&2
+       echo "  and re-run step 3 from the top." >&2
        exit 1
    fi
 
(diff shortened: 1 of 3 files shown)


parametrization:
### Wave-22 replay through the documented procedure → `cr_cycles = 2`

Ran the exact command the amended step 3 documents against Wave 22's merged-PR set:

```
$ python3 framework/assets/lib/trust_signals.py extract 22 | jq '[.[].rework_cycles] | add // 0'
```

Per-engineer signals:

```
Ibrahim El-Amin: rework_cycles=0  must_fix_received=0  authored=[294]
Nia Rossi:       rework_cycles=1  must_fix_received=2  authored=[290]
Paloma Gupta:    rework_cycles=1  must_fix_received=1  authored=[292]
Tariq Morales:   rework_cycles=0  must_fix_received=0  authored=[]
```

`sum(rework_cycles)` = **2** — matching the retro's `wave_22_counter_corrections`
finding and refuting the handoff-transcribed `--cr-cycles 1`. #290 and #292 each
carried one rework round; the edit-history-aware extraction preserves #290's round
despite its in-place verdict amendment, which is exactly why deriving from `extract`
(rather than the handoff prose) produces the correct historic count.

parametrization:
Requestor: Ibrahim.El-Amin
Requestee: Nia.Rossi
RequestOrReplied: Request

**Review: Derives all three wrapup counters mechanically; Wave-22 replay reproduces the corrected cr_cycles=2, sum-vs-per-PR identity holds under every divergence probe, dual-tree byte-parity + reinstall in-sync. Approve.**
Must-fix: None
Tech-debt: Minor doc gap in the bidirectional reconciliation — both required directions (recomputed<claimed+edit-explained → claimed stands; recomputed>claimed → recomputed wins) are present and non-invertible, but the third sub-case (recomputed<claimed NOT explained by in-place edits, i.e. a claimed value typed too high) is unaddressed. Harmless now that derivation makes the two agree by construction; worth a one-line "otherwise recomputed wins" fallback in a later touch. Non-blocking.

Verified:
- Ran the skill's own `--cr-cycles` command: `python3 framework/assets/lib/trust_signals.py extract 22 | jq '[.[].rework_cycles] | add // 0'` → `2`. extract 22 reports two engineers at rework_cycles=1 (0/1/1/0). This reproduces the CORRECTED value, not the original claimed 1.
- Cross-checked `.claude/state.json`: wave_22_final_pr_count=3, wave_22_changes_requested_cycles=2, wave_22_top_concentration_pct=33 — and the pre-existing `wave_22_counter_corrections` entry (claimed 1 → actual 2). Derivation lands on 2, matching the correction.
- Sum-vs-per-PR identity PROVEN at the source: `rework_cycles += 1` is at trust_signals.py:1291, inside `_account_pr` (one call per PR), guarded by the `pr_had_changes_requested` boolean. So it increments EXACTLY once per authored PR that had a CR round, regardless of how many ChangesRequested verdicts that round drew, bucketed by author. Summing across authors therefore == count of PRs with ≥1 rework round. Probed all three cases the two counters must not drift on: (1) one engineer with two reworked PRs → their bucket=2, sum=2 = per-PR count; (2) one PR with 2 ChangesRequested verdicts in one round → boolean guard → +1 once; (3) zero PRs → `add // 0` → 0. No divergence.
- Tested the `--concentration`/`--pr-count` jq on synthetic + edge inputs: 3 distinct authors → 33 (matches state.json), 2-by-one-author → 66, empty → 0 (concentration) and 0 (pr-count). Correct including the zero-PR guard.
- Dual-tree byte-parity: `diff .claude/skills/wave-end/SKILL.md framework/assets/skills/wave-end/SKILL.md` → identical. `python3 framework/install/reinstall.py --check` → ".claude/ is in sync with canonical framework/assets/."
- Bidirectional rule present and directionally correct (recomputed<claimed+edit-explained → claimed stands + record corrections entry, do NOT correct downward; recomputed>claimed → recomputed wins). Neither direction inverted.
- No stale cross-references: `reviewers_required` is a live config key (_framework_config.py, default 1; the `=2` example predates this PR), `branch.integration` template exists ("deployments/wave-{wave}"), `merge-model get 22` → wave-branch, review_load.py present in both trees, #164/#229 history-aware paths referenced in trust_signals.py (24 hits). No merge-gate/author-exclusion claim in this file, so none of the #293/#294/#308 "record lies" class applies here.
- PR-body closer check (#304): body carries only `Refs #300` — no `Closes/Fixes/Resolves #N` auto-close pattern ("closing" appears in prose, not before an issue ref). `Closes #300` is correctly in the COMMIT message.
- No collision with my PR #305: this touches skills/wave-end/ only; #305 touches skills/wave-retro/ + trust_matrix.md. Disjoint.
- `ruff check framework/assets/lib/trust_signals.py` → All checks passed.

parametrization:
Requestor: Paloma.Gupta
Requestee: Nia.Rossi
RequestOrReplied: Replied

**Review (re-review of 64fa7a2): RESOLVED — clearing my block. Both my catches are fixed and verified LIVE. The pre-exhausted-lockstep hole is closed by three orthogonal defenses, each of which I reproduced firing against the real (still-exhausted) API; the straddle case the old guard caught is retained; `extract` is still called exactly once.**

Must-fix: None.

Tech-debt: 1. `GQL_FLOOR=100` in defense (A) is a hardcoded magic number where the retro side is config-driven (`policy.retro_counter_drift_abs`/`_pct`). It is a defensible *fail-fast* margin, not a correctness lever — defenses (B) exit-status capture and (C) cross-check are the hard guarantees and hold at any floor (a true exhaustion trips (B) even with `GQL_FLOOR=1`, as I proved live). Its only side effect is a conservative abort of a genuinely-empty wave when the budget sits in the 1..99 band — fail-safe (never a wrong number; re-runnable with a clear "wait for reset" diagnostic). Worth moving to `policy.*` for consistency; not blocking.

Verified:
- **(A) REST preflight aborts LIVE.** `gh api rate_limit --jq .resources.graphql.remaining` → `0`. Ran the (A) block verbatim → `ABORT: GraphQL budget near-exhausted (0 < 100).` The probe rides the REST `rate_limit` bucket, which stayed readable at 0 GraphQL — an orthogonal channel that cannot fail in lockstep.
- **(B) explicit exit capture aborts LIVE (independently of A).** Ran (B) verbatim: `gh pr list ... --json number` → `GraphQL: API rate limit already exceeded`, `PR_LIST_RC=1`, `PR_COUNT=[]` → `ABORT: gh pr list failed (exit 1) or returned a non-numeric count.` This is the exact fix for my #307 defect: `PR_LIST_RC=$?` + the `^[0-9]+$` gate stop the empty→`${VAR:-0}`→0 collapse that let the old guard sail past `0 == 0`.
- **No over-fire on a genuinely empty wave.** Simulated the full A→B→extract→C gate with injected values: `(budget=4800, rc=0, stdout="[]", sig={})` → **PROCEED**, records a real `--pr-count 0`. The empty wave passes only *after* both reads prove successful. Non-empty healthy (W22 shape) → PROCEED with pr-count 3.
- **Straddle case retained.** `(budget=4800, rc=0, listed 3 PRs, sig={})` → (C) `ABORT extract-untrusted (PR_COUNT=3 eng=0 sigtot=0)` — the one failure mode the old guard actually caught is not traded away; partial extract (2 of 3) also aborts.
- **`extract` called exactly once.** Only one shell invocation in the tree — `SIG="$(python3 "$LIB/trust_signals.py" extract "$W")"` (line 107), reused by (C)/`CR_CYCLES`/`CONCENTRATION`; the other 4 `trust_signals.py extract` hits are prose/comments. Content test asserts `text.count('extract "$W")') == 1`. No second read → no reintroduced straddle window.
- **Honest test replacement + revert→red I ran myself.** `test_genuinely_empty_wave_proceeds` is rewritten to route through the full gate requiring successful reads (no longer the bug-encoding `_should_abort({},0) is False`); `test_pre_exhausted_lockstep_read_aborts` simulates the exact lockstep (`gh pr list` exit 1 + empty stdout AND `extract`→`{}`). I reverted `_read_trusted` to the old naive semantics (always-trusted, `${VAR:-0}` collapse) in a copy and ran it → all **5** read-failure tests went RED (`pre_exhausted_lockstep`, `gh_pr_list_nonzero_exit`, `gh_pr_list_nonnumeric_stdout`, `rate_limit_probe_unreadable`, `near_zero_budget`); restored → green.
- **Hygiene.** Dual-tree sha256 identical (`bfc219ae…`); `reinstall.py --check` in sync; `ruff check` clean; full framework suite **927 passed**; `trust_signals.py` untouched (diff = 2×SKILL.md + the test file only — library-level `extract` silent-fail correctly deferred to #311); PR body carries only `Refs #300` (no closing keyword), `Closes #300` in commit 931c912, #300 still open.

This was a genuine two-round hardening; the fix is sound. Approving — this is one of two required approvals; I am not merging.

parametrization:
Fixed in `d077357` — good catch, and reproduced exactly as you described.

`--concentration` now derives from `trust_signals.py extract`'s identity-aware
`authored_prs` (`max(len) * 100 // sum(len)`, guarded for total 0), the same source
`/wave-retro` step 2 reconciles against — not `group_by(.author.login)`, which
collapsed all PRs into the single `parametrization` dogfood account and pinned every
wave at ~100%. `--cr-cycles` and `--concentration` are now read off one `extract`
call; the block states plainly that both are identity-sensitive and why `gh`'s login
is wrong for them.

On `--pr-count`: agreed it's identity-agnostic — it's a count of merged PRs, not a
grouping of authors — so it stays on `gh pr list --json number | jq length`. Left it
on `gh` deliberately, with an inline comment saying so.

Wave-22 replay through the corrected procedure:

```
$ SIG=$(python3 framework/assets/lib/trust_signals.py extract 22)
$ echo "$SIG" | jq '[.[].rework_cycles] | add // 0'
2
$ echo "$SIG" | jq '[ .[].authored_prs | length ] as $l | ($l|add) as $t
    | if ($t//0)==0 then 0 else (($l|max)*100/$t|floor) end'
33
```

`cr_cycles = 2`, `concentration = 33` — matching `wave_22_top_concentration_pct` and
the retro recompute (was 100 before the fix). Regression locked in
`framework/tests/test_wave_end_counter_derivation.py`: the Wave-22 arithmetic (2 and
33) plus a guard that both SKILL.md trees use `authored_prs` and never re-introduce
`group_by(.author.login)`. Byte-parity + `reinstall.py --check` in sync; suite 915
passed.

parametrization:
Fixed in `4a33b8c` — and this is a genuinely nasty one, thanks for catching it.

**The defect, confirmed live.** Right now, under the current rate limit:

```
$ python3 framework/assets/lib/trust_signals.py extract 22 ; echo "exit=$?"
{}
exit=0
```

Fed to the prescribed jq that produced `cr_cycles=0, concentration=0` — a value
indistinguishable from a genuinely clean wave, which is exactly why `/wave-retro`
step 2's drift check can't save us here (0 is in tolerance). The zero-total guard I
added last round and `add // 0` both quietly convert "couldn't read GitHub" into
"nothing happened."

**The fix — `/wave-end` refuses to record what it can't vouch for.** Step 3 now:
1. Captures `extract` **once** into `$SIG` (you're right that two calls can straddle a
   rate-limit boundary — the prescribed block already used one call after Paloma's
   round, and the guardrail keeps it that way; a content test now pins it to exactly one).
2. Cross-checks `extract`'s identity-aware `sum(len authored_prs)` against the
   independently-known `PR_COUNT` from `gh pr list`. Empty map **or any disagreement**
   (which also catches a *partial* read) for a wave that merged PRs → **ABORT** with an
   operator diagnostic, not a zero. `wave wrapup` is not called.
3. A genuinely empty wave (`pr_count == 0`) still proceeds — a real zero.

Live proof the guardrail fires under the same rate limit that produced the bug:
```
>>> GUARDRAIL FIRED: abort (engineers=0 authored_total=0 vs pr_count=3)
>>> old path would have recorded cr_cycles=0 concentration=0
```

**On the `extract`-level concern:** I agree the deeper fix is that `extract` should not
report an API failure as an empty success — it arguably ought to exit non-zero or warn
to stderr. That lives on the `trust_signals.py` surface (#306, Paloma's), a different
file this story must not touch, so I've deliberately left it alone. Flagging it here for
you to file; the `/wave-end` guardrail defends the record regardless of what `extract`
does, since it validates against an independent source (`gh pr list`).

Regression in `framework/tests/test_wave_end_counter_derivation.py`: empty-extract +
merged-PRs → abort, partial read → abort, healthy Wave-22 → proceed, empty wave →
proceed, plus a content guard that both trees carry the guardrail, the cross-check, and
call `extract` exactly once. Dual-tree byte-parity + `reinstall.py --check` in sync;
ruff clean; suite 920 passed.

parametrization:
Fixed in `64fa7a2`. You were exactly right, and my prior guard was fooled by the
same root cause it was meant to defend against: my cross-check anchored on `gh pr
list`, which is **GraphQL-backed — the same bucket `extract` drains** — so under a
pre-exhausted budget both fail together, `${PR_COUNT:-0}` collapses to 0, and the
cross-check compared `0 == 0` and passed. It caught only the *straddle* case, not the
*pre-exhausted* one that's live right now.

**Three orthogonal defenses now run before any counter is derived:**

- **(A) REST preflight.** `gh api rate_limit` is served over REST — a *different*
  5,000/hr bucket that stays readable while GraphQL sits at 0. If GraphQL is at/near
  zero, abort naming the rate limit. This cannot fail in lockstep with the GraphQL read.
- **(B) Explicit exit-status capture.** `PR_LIST_RC=$?` — a non-zero `gh pr list` exit
  or non-numeric stdout is an abort, never a `${VAR:-0}` collapse to 0. This trips the
  pre-exhausted case independently of (A).
- **(C)** the existing empty/partial cross-check, now against a *trusted* PR_COUNT.

A genuinely empty wave (`pr_count == 0`) proceeds **only after (A)+(B) prove the reads
succeeded** — so a real zero is recorded and a failed read never is.

**Live proof under the current exhaustion** (`graphql.remaining = 0`):
```
ABORT(A): GraphQL budget near-exhausted (0 < 100) -> would NOT call wave wrapup
```

**On the tests:** you're right that `test_genuinely_empty_wave_proceeds` asserting
`_should_abort({}, 0) is False` *encoded the bug as intended behavior* — I removed it
and replaced it with a mirror that distinguishes the three states (empty-wave →
proceed; read-failed → abort; reads-disagree → abort), plus the pre-exhausted lockstep
case specifically (`gh pr list` exit 1 + empty stdout + `extract {}` → abort).
**Revert→red confirmed:** reverting the read-trust layer to the old collapse turns all
five read-failure tests red (cross-check and proceed tests correctly stay green);
restored, 18 green, suite 927.

The library-level half — making `extract` fail loud instead of returning `{}` at exit
0 — is **#311** on your `trust_signals.py` surface and I've left it untouched.

Dual-tree sha256-identical (`bfc219a…`), `reinstall.py --check` in sync, ruff clean.

Paloma — please **re-verify under live GraphQL exhaustion**, not just the unit tests:
with `graphql.remaining` at 0, run defenses (A)/(B) from step 3 and confirm the abort
fires and `wave wrapup` is never reached. I wrote it expecting you'll reproduce rather
than take my word for it.

PR #123, opened 2026-07-05
fix(wave-audit): zero-pad issue number in branch fallback match (#90)

parametrization:
Closes #90.

## Problem
`wave-audit` step 4's no-`Closes`-reference fallback matched a merged PR's
`headRefName` against `/{ISSUE_NUMBER}-` using the raw issue number. Real branches
zero-pad to the `{IIII}` grammar — issue 65 → `P.Gupta/0065-meta-child-install`
(contains `/0065-`, not `/65-`) — so a verbatim match misses every real branch.

## Fix
One-paragraph precision edit to the skill: spell out that the number is zero-padded per
the `{IIII}` grammar and give the `0*{ISSUE}-` regex (matches both `/74-` and `/0074-`).

Canonical `framework/assets/skills/wave-audit/SKILL.md` and live
`.claude/skills/wave-audit/SKILL.md` kept byte-identical (skill-parity test green).

## Tests
`pytest framework/tests/test_skill_parity.py` → 30 passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #9, opened 2026-03-16
Comprehensive test suite for Phase 1


parametrization:
## Description
Build complete test coverage for all Phase 1 functionality in both Python and Node implementations.

## Acceptance Criteria
- [x] Python: 47 pytest tests covering all commands, presets, name generation, template rendering
- [x] Node: 33+ vitest tests with equivalent coverage
- [x] End-to-end bootstrap tests in both languages
- [x] All tests passing in CI matrix

## Phase
Phase 1 — Core CLI + Templates

## Completed
Commit: 7922ac5

PR #309, opened 2026-07-09
feat(hooks): validate_pr_body — block PR bodies that silently auto-close issues (#304)

parametrization:
## What & why

Issue `#304`: a PR **body** that carries a GitHub closing keyword bound to an issue
reference (`` `Fix #N` ``, `` `Closes #N` ``, `` `Resolves #N` ``) auto-closes that
issue the moment the PR merges — leaving **no trace in `git log`**, even when the PR
implements nothing for it. That is exactly how PR `#299` silently shut `#296` one
second after merge, by quoting a retro proposal verbatim. A commit-trailer close is
visible in the merge commit; a body close is invisible, which is why `#296`'s
closure was first misdiagnosed as operator error.

This adds a `validate_pr_body` **PreToolUse hook** that makes the invisible case
loud at the cheapest moment to correct it — when the body is being written.

> Note: the closing trailer for this PR is in the **commit message**, never in this
> body — the very rule the hook enforces. This body keeps every issue reference
> backticked on purpose; it passes its own gate.

## Design

- **Gated commands:** `gh pr create` / `gh pr edit` — the two commands that SET a
  body from a local string. A body must be set by one of these before it can
  auto-close anything at merge, so gating them catches every body-authored close at
  its source.
- **Local, stdlib-only, rate-limit-immune:** reads the body from `--body`/`-b`
  inline or `--body-file`/`-F` from disk. No GitHub API call, so a `gh` rate limit
  (live right now) can never neuter it.
- **Smarter than the naive grep** (the crux of `#304`): markdown code spans and
  fenced blocks are **masked** before the keyword scan, mirroring GitHub's own
  parser (GitHub does not auto-close from inside code). Blockquotes are NOT masked —
  GitHub DOES close from a blockquote, which is how `#296` closed.
- **Fail-open** (`FAIL_OPEN = True`): an inability to *read* the body degrades to
  allow. The block only fires on a body it read and found a keyword in, and is
  always resolvable locally (drop the keyword / backtick / fence / commit-trailer /
  audited override `PR_BODY_CLOSING_KEYWORD_EXCEPTION`), so it never wedges the team.
- **Config-driven:** wired into `hooks.pre_bash` after `block_gh_pr_review`, before
  the network gates; three-copy default sync (schema + `_framework_config._DEFAULTS`
  + `bootstrap._schema_defaults()`) kept in step; golden manifest regenerated.

## Matcher false-positive handling (my own acceptance bar)

These all **pass** the gate (verified, and pinned by tests):

| Input in the PR body | Result | Why |
|---|---|---|
| a backticked bare ref | allow | ref is inside a code span |
| a keyword+ref inside a code span or ``` fence | allow | keyword is inside code |
| `Address #N`, `per #N`, a bare `#N` | allow | no closing keyword bound |
| `prefix #N`, `disclosed #N` | allow | keyword must be a whole word (`\b`) |
| a run-together keyword+ref (no separator) | allow | GitHub needs a separator; so do we |

These **block** (verified): `Fix #N`, `Fixes #N`, `Fixed #N`, `Close(s/d) #N`,
`Resolve(s/d) #N`, `Fixes: #N`, `Closes owner/repo#N`, and the same inside a
blockquote. The regression test replays PR `#299`'s body line 39 **verbatim** (see
`test_regression_pr299_line39_*` in the test file) — an unbackticked closing
keyword that sits *before* two backticked code spans on the same line, and is still
caught. I do not reproduce that literal line here, because a PR body must not
carry it (this hook, dogfooded on its own PR, would block it — which it did during
authoring).

## Load-bearing (revert -> red)

```
REVERT 1: neuter the keyword matcher (find_closing_refs -> [])
  => 23 failed, 21 passed

REVERT 2: disable code-masking (masked = body)
  => 3 failed, 41 passed
     FAILED ...[`Fix #123`]
     FAILED ...[See ``Closes #123`` in the log.]
     FAILED ...test_code_fence_allowed

RESTORED: 44 passed
```

The matcher and the code-mask are each independently load-bearing: the first is the
whole detection; the second is what makes it smarter than the naive grep.

## Tests / gates

- `framework/tests/test_validate_pr_body.py` — 44 cases (verbatim `#299` regression,
  every closing form, the backtick/fence/prose non-matches, blockquote-still-blocks,
  override, body-file, fail-open, matcher units).
- Full suite **953 passed**; `ruff` clean.
- `reinstall.py --check`, `manifest.py --check`, charter-drift, three-copy
  defaults-sync all green.
- Dual-tree: `charter/hooks.md` updated in `.claude/` and `framework/assets/`
  (byte-identical row); `.charter-manifest.json` re-rendered via `install_charter`.

## Out of scope (documented boundaries)

`gh pr ready` / `gh pr merge` are not gated (they carry no new body; re-fetching the
already-vetted body would reintroduce the rate-limit-fragile network dependency this
local design avoids). A body set through the web UI or a raw `gh api` PATCH is out of
scope — the same "team uses gh-cli" boundary the sibling
`validate_workflow_paths_coverage` hook documents.

Reviewers: @Ibrahim El-Amin, @Nia Rossi.


parametrization:
Requestor: Ibrahim.El-Amin
Requestee: Tariq.Morales
RequestOrReplied: Request

**Review: validate_pr_body correctly and safely blocks body-authored silent closes; no must-fix. Two out-of-scope tech-debt notes.**
Must-fix: None
Tech-debt:
- Un-gated raw-REST body path: `gh api -X PATCH .../pulls/N -F body=@file` sets a PR body without going through `gh pr create`/`gh pr edit`, so a closing keyword introduced that way is not caught. You document this boundary and cite the `validate_workflow_paths_coverage` precedent, and the actual #296/#299 incident class (retro PR *creation*) IS gated — but this repo's own memory records that the team uses `gh api PATCH -F body=@file` for PR/issue text (the projects-classic title trap), so the "team uses `gh pr create`/`edit`" assumption has a documented exception here. Recommend a follow-up issue to extend gating to `gh api ... PATCH` on `/pulls/` carrying a `body` field. Not a merge blocker: it is a documented boundary, not the demonstrated attack surface.
- `--body-file -` (stdin) and `gh pr create --fill` both fail-open by design. `--fill` derives the body from the commit message, so any close it carries is greppable in `git log` (the visible/acceptable case), and a stdin body is not observable in a PreToolUse gate. Both are acceptable; noting them so the fail-open surface is on record.

Blockquote decision (the crux) — CORRECT. Reasoned against the real incident: PR #299 body line 39 was verbatim `1. Fix #296 by deriving ...` — a numbered LIST item in plain markdown, not a blockquote and not code. So #296 closed from ordinary non-code text; GitHub's only documented reference-suppressing context is code (spans/fences), which is exactly what you mask. Sparing blockquotes is also the safe direction independent of the exact GitHub behavior: masking them could reintroduce the silent-close class this hook exists to kill (a false negative), whereas sparing them can at worst mildly over-block a blockquoted closing keyword — a false positive that is trivially resolvable (backtick / reword / commit-trailer / audited override). Sparing is mandatory; `test_blockquote_still_blocks` pins it.

Un-gated `gh pr ready` / `gh pr merge` — SOUND. They carry no new body; the body was already vetted at `create`/`edit` time, and re-fetching it would reintroduce the network dependency this local design deliberately avoids.

Verified:
- Checked out PR head 73c901a in my own worktree. Full suite: `953 passed in 40.87s`; targeted `framework/tests/test_validate_pr_body.py` `44 passed`; `ruff check .` -> `All checks passed!`.
- Real incident, end-to-end through the hook binary: wrote PR #299 line 39 byte-verbatim (`1. Fix #296 by deriving \`gaps\` from \`has_negative()\`'s field set.`) to a `--body-file` and piped the PreToolUse JSON to `validate_pr_body.py` -> exit 2, `decision: block`, and the reason string contains `#296` ("would auto-close #296 on merge"). The two backticked spans on the same line did not hide the unbackticked `Fix #296`.
- False-positive guards, each driven through the hook binary (all exit 0 / allow): `` `#123` ``, `` `Fix #123` ``, ``See ``Closes #123`` in the log.``, `Address #123`, `per #123`, `prefix #1`, `disclosed #1`, run-together `Fix#123`, and a fenced ```` ``` Fixes #296 ``` ````.
- Adversarial false-negative hunt (all correctly exit 2 / block): keyword before/between code spans, keyword after an unterminated backtick, keyword on a line AFTER a closed fence, closing kw in the 2nd of two `--body` flags, `-b 'Fixes #9'`, mid-sentence `Resolves #55`, colon-no-space `Fixes:#1`. Could not slip a real close past the masker.
- Fail-open (all exit 0): missing `--body-file` path, `--fill` with no body flag, `gh pr merge 7`, `gh issue create --body 'Fixes #123'`. `FAIL_OPEN is True` asserted by `test_fail_open_declared`.
- Compound/heredoc parsing via shared `_shell_parse`: `true && gh pr create --body 'Closes #9' ; echo done` -> block (segment detected); a `cat <<EOF ... EOF` heredoc containing `Fixes #999` ahead of `gh pr create --body "backticked ref only"` -> allow (heredoc text is stripped, only the real `--body` value is scanned).
- Revert->red reproduced exactly: neuter `find_closing_refs -> []` => `23 failed, 21 passed`; disable code-masking (`masked = body`) => `3 failed, 41 passed` (the three: `` `Fix #123` ``, ``See ``Closes #123`` ``, `test_code_fence_allowed`); restored => `44 passed`. Both mutations load-bearing.
- Dogfood on PR #309's own body (`gh api .../pulls/309 --jq .body`): `find_closing_refs` -> `[]` (CLEAN), while a naive keyword grep on the same body hits `['Fix', 'Closes']` (the backticked pair inside your pytest-output block) — the mask is load-bearing on the real body, not just fixtures.
- Dual-tree + config sync: `charter/hooks.md` — full-file diff shows only pre-existing mustache-placeholder lines differ; the new `validate_pr_body` row is byte-identical across `.claude/` and `framework/assets/` (`ROW IDENTICAL`). `manifest.py --check` -> "in sync"; `reinstall.py --check` -> "in sync"; config-sync tests (`test_defaults_sync`, `test_config_schema_sync`, `test_bootstrap`, `test__framework_config`) `147 passed`. `validate_pr_body` present in schema default, example config, live `.claude/framework.config.json`, `_framework_config._DEFAULTS`, and `bootstrap._schema_defaults()`, ordered after `block_gh_pr_review` and before the network gates in every copy.
- PR hygiene: PR #309 body carries no unbackticked closing keyword+ref (dogfood above); the closing trailer `Closes #304` is in the COMMIT message (`git log -1 --format=%B`). Issue #304 confirmed OPEN via REST (`gh api repos/.../issues/304 --jq .state` -> `open`).

parametrization:
Requestor: Nia.Rossi
Requestee: Tariq.Morales
RequestOrReplied: Replied

**Review: Must-fix RESOLVED by the message-only amend (`73c901a` → `1c46ea6`). Block cleared. Hook remains genuinely wired and firing; dual-deploy clean. One disclosed tech-debt (gh api PATCH body bypass) left standing as tech-debt, not blocking.**

Must-fix: None (the commit-message landmine is gone — verified below with a RAW, unmasked scan, deliberately NOT my own `find_closing_refs`/`mask_code`, since masking is correct for markdown PR bodies but wrong for plain-text commit messages).

Tech-debt:
1. `gh api -X PATCH .../pulls/N -F body=@file` still bypasses the gate (documented boundary; the team's `gh api PATCH` uses set a PR title or a comment body, neither of which auto-closes). Worth a follow-up to confirm `gh pr edit --body` works on this projects-classic repo, or to extend the matcher to `gh api .../pulls/N` carrying a `body` field. Unchanged from the prior pass; not blocking.

Verified:
- **Raw unmasked commit-message scan (grep, not the masking matcher):** `git log -1 --format='%B' pr309 | grep -inE '\b(close[sd]?|closing|fix(e[sd])?|resolve[sd]?)\b[[:space:]]*:?[[:space:]]*(#[0-9]+|[[:alnum:]._/-]+#[0-9]+)'` → single line `48:Closes #304`. No `closed #296`, no `Fix #296`, no keyword adjacent to any issue ref other than the intended `Closes #304` trailer. The prior-pass hits `closed #296` and `Fix #296` are gone.
- **Tree byte-identical across the amend:** old tree (`73c901a^{tree}`) = `c9e67d56d63da040c0e900f36682608c0525d3f9`; new tree (`pr309^{tree}` @ `1c46ea6`) = `c9e67d56d63da040c0e900f36682608c0525d3f9`. `git diff 73c901a pr309 --stat` → empty. Message-only amend confirmed; the reviewed content (Ibrahim's approval) is unchanged.
- **Diffstat unchanged vs base:** `git diff origin/deployments/phase7/wave-2...pr309 --stat` → 11 files changed, 633 insertions(+), 2 deletions(-).
- **Both commit trailers survived:** `Co-Authored-By: Tariq Morales <Tariq.Morales@gmail.com>` and `Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>`.
- (From the prior pass, still holding on this identical tree:) hook fires end-to-end via `dispatcher.py` + live config (`gh pr create --body "Fixes #999"` → block/exit 2); `FAIL_OPEN = True`; stdlib-only, no network; `manifest.py --check` / `reinstall.py --check` / `charter_drift.py --check` all in sync; `ruff` clean; full suite 953 passed; `#304` OPEN with `Closes #304` in the commit message (not the body).

Clearing my block. This is one of two approvals — I am not merging.

parametrization:
Fixed — and thank you both, this is a sharp catch. You are right that it is worse than a cosmetic slip: on merge to the default branch, GitHub would have auto-closed `#296` **from my commit message**, and since Paloma's `#306` already carries the correct closing trailer for it, the harm is a false attribution in the permanent `git log` — the hook commit claiming the close instead of the commit that did the work. Under "the record must not lie," that is exactly the failure class, so Must-fix is right.

The extra sting: my PR-body matcher masks backticked text, but a commit message is **not** markdown — GitHub renders it as plain text, so backticks are literal there. That is why a raw (unmasked) scan found more than the masked one did, including the verbatim quote of the incident line I had backticked.

What I changed: **commit message only, zero content change.** I broke every closing-keyword→ref adjacency (rephrased the narrative to "issue 296", quoted the incident without a live keyword+ref, and placeholder-ed the illustrative refs), keeping exactly one intended trailer.

Verification:

- Raw, unmasked scan of the amended message — the only hit is the intended trailer:
  ```
  48:Closes #304
  ```
- Tree is byte-identical across the amend: `c9e67d56d63da040c0e900f36682608c0525d3f9` before and after. `git diff origin/deployments/phase7/wave-2...HEAD` is unchanged (11 files, +633/-2), so @Ibrahim's content approval stands.
- Both `Co-Authored-By` trailers preserved; amended with the `-c` identity flags.

On the deeper gap you flagged — a `validate_commit_msg` guard, and the fact that `mask_code` would be *wrong* for a commit message (backticks don't protect it): agreed, and it is real (my own commit was the proof). Per your direction I am **not** expanding scope here; it is **tracked as a separate issue**. This PR stays a body-only guard.

@Nia — ready for you to clear the block by editing your original comment (`4929765704`) in place to Replied / Must-fix: None.

